PR #3, opened 2025-11-08
Feat/logs

yuzurihaaa:
A start for implementing Logs functionality. PR is not a fully ready state since I am working on SDK part, but would appreciate some review / feedback.

Patterns are referenced based on existing code.


michaelbushe:
The start for implementing Logs must be done in the API project, as referenced in the Issue. 

It must be documented and tested and work in no op in the API and comply with the specification.

yuzurihaaa:
Hi @michaelbushe 

Yep, I am aware of it. The API here is following https://opentelemetry.io/docs/specs/otel/logs/api/ for `LoggerProvider` and `Logger`, that's pretty much it afaik :)

I am used to verifying stuff end to end before creating a PR, but since the API is in different repo. I decided to create a PR for early review in case if there's anything that I missed out, patterns and stuff so that I can fix them while working on the SDK codes. I'd consider this as done after I verified on SDK that everything is working e2e. Which... might take a while.

example/opentelemetry_api_example.dart
```diff
@@ -63,4 +63,8 @@ void main() {
       OTelAPI.attributes([OTelAPI.attributeString('event-foo', 'bar')]));
   rootSpan.end(
       spanStatus: SpanStatusCode.Ok); //Capitalized Ok to match the OTel spec
+
+  final defaultGlobalAPINOOPLoggerProvider = OTelAPI.loggerProvider();
+  final logger = defaultGlobalAPINOOPLoggerProvider.getLogger('dart-otel-api-example-service');
+  logger.emit(body: "some log here!", attributes: equalToTheAbove, context: Context.current);
```
michaelbushe: Remove the context: param, it's not needed and examples are important because people copy them

lib/dartastic_opentelemetry_api.dart
```diff
@@ -65,3 +65,7 @@ export 'src/api/trace/tracer_provider.dart';
 export 'src/factory/otel_factory.dart';
 // Util
 export 'src/util/otel_log.dart';
+// Log
```
michaelbushe: These need to be in alphabetical order.

lib/src/api/factory/otel_api_factory.dart
```diff
@@ -114,6 +115,19 @@ class OTelAPIFactory extends OTelFactory {
         serviceVersion: serviceVersion);
   }
 
+  @override
+  APILoggerProvider loggerProvider({
```
michaelbushe, line 124: Needs comments.  Comments are inconsistent in this file but we should do aim high.

lib/src/api/logs/logger.dart
```diff
@@ -0,0 +1,54 @@
...
+    this.version,
+    this.attributes,
+  });
+
+  /// Returns true if the logger is enabled.
+  bool get enabled => false;
+
+  /// Emit a LogRecord.
+  ///
+  /// More info https://opentelemetry.io/docs/specs/otel/logs/api/#emit-a-logrecord
```
michaelbushe, line 42: 👍

lib/src/api/logs/log_record.dart
```diff
@@ -0,0 +1,27 @@
+enum SeverityNumber {
```
michaelbushe: Please rename this file to severity.dart. 

We need a second enum, LogLevel (don't reuse OTelLog's), that corresponds to SeverityText.  TRACE, DEBUG,etc.  These should have values for comparison and a method for fromSeverityNumber

"Description: severity text (also known as log level). This is the original string representation of the severity as it is known at the source. If this field is missing and SeverityNumber is present then the short name that corresponds to the SeverityNumber may be used as a substitution. This field is optional.
"

Needs comments from:
https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-severitynumber

Instead of just names, these must have an integer value for SeverityNumber (0-24) so they can be compared. We'll need this when we integrate different loggers. See:
https://opentelemetry.io/docs/specs/otel/logs/data-model/#comparing-severity

They also need and ALL CAPS Short name like "TRACE2".   

Though Dart uses lower case (or camelCase) enums, the specification uses ALL CAPS for the enum names.  I've taken the pattern of following the spec (For example, SpanStatus.Ok - which is neither camelCase nor All CAPS but follows the spec).  Please change these to TRACE2, DEBUG3, etc.  It makes more sense since it's common for logging. 

Furthermore, these need toString overrides and

lib/src/api/logs/logger_provider.dart
```diff
@@ -0,0 +1,211 @@
...
+
+part 'logger_provider_create.dart';
+
+/// APILoggerProvider is the entry point of the OpenTelemetry logger API.
+/// The API prefix indicates that it's part of the API and not the SDK
+/// and generally should not be used since an API without an SDK is a noop.
+/// Use the LoggerProvider from the SDK instead.
+/// It provides access to [APILogger]s which are used to log operations.
+/// You cannot create a LoggerProvider directly;
+/// you must use [OTelAPI] or more likely [OTel], for example to get the default tracer:
```
michaelbushe: default logger, not tracer
Also, this uncovers an error in tracer doc too.  Both these lines only return loggerProviders.  This should have a second part: 
final (not var) logger = loggerProvider.getLogger()
final tracer = tracerProvider.getTracer()

lib/src/api/logs/logger_provider.dart
```diff
@@ -0,0 +1,211 @@
...
+    String? effectiveVersion = version;
+    String? effectiveSchemaUrl = schemaUrl;
+
+    // Only apply defaults if all optional parameters are missing
+    if (version == null && schemaUrl == null && attributes == null) {
+      effectiveVersion = OTelAPI.defaultServiceVersion;
+      effectiveSchemaUrl = OTelAPI.defaultSchemaUrl;
+    }
+
+    final key = _LogKey(
```
michaelbushe: Add    // Create a cache key based on the provided parameters.
If you going to copy code (which is ok), copy comments too. :)

lib/src/api/logs/logger_provider.dart
```diff
@@ -0,0 +1,211 @@
...
+
+      return true;
+    } catch (e) {
+      OTelLog.error('Error during LogProvider shutdown: $e');
+      return false;
+    }
+  }
+}
+
+class _LogKey {
```
michaelbushe: I see you made the mistake I did: I copied the TracerKey to a MetricKey, then you copied it to a LoggerKey. If you want, remove them all and replace with one SignalInstanceKey, or you can leave me to do it since I made the mistake first. :)

README.md
```diff
@@ -220,6 +220,16 @@ void main() {
 }
```
michaelbushe, line 221: Please change "Coming soon - Logs" to "Log" !

README.md
```diff
@@ -220,6 +220,16 @@ void main() {
...
 ```
 
+### Working with logging
+```dart
+import 'package:dartastic_opentelemetry_api/dartastic_opentelemetry_api.dart';
+
+final loggerProvider = OTelAPI.loggerProvider();
+final logger = loggerProvider.getLogger('dart-otel-api-example-service');
+logger.emit(body: "some log here!", attributes: equalToTheAbove, context: Context.current);
```
michaelbushe: Fussy but to keep the same serious example "voice" use either "foo occurred" or maybe "User foo did something".  

You can't use equal to the above - it's a new box and that name was to show equivalence between two ways of making attributes.  Examples are very important - if you keep this guaranteed you will see "equalToAbove" in real code or in articles, maybe nowadays in an LLM.

Instead, make this very real and something people should copy.  Remove context:, it's not needed.  

Show a few examples for emit that cover all the variables (maybe not timestamps). Include different attributes (use Attributes.of() and/or .toAttributes() to be terse.  Show usages of the severity level enum and then severityText/logLevel.

Lastly, since body is dynamic, show an example with an array, we'll at least have to handle that in the SDK, probably should handle anything that's a toString.

void emit({
    DateTime? timeStamp,
    DateTime? observedTimestamp,
    Context? context,
    SeverityNumber? severityNumber,
    String? severityText,
    dynamic body,
    Attributes? attributes,
    String? eventName,
  }) {

michaelbushe (commented):
Overall this is a good submission, thank you.  The tests all pass but the coverage (./tool/coverage.sh) shows only 57.7 % coverage.  Please try to pull that up to 80%.  I'll try to pull up the util and common coverage to 80%. They are higher than logs but are missing coverage for recent changes.

lib/src/api/logs/logger_provider.dart
yuzurihaaa: Updated (manually marking this one by comment as the changes are not marked as outdated)

lib/src/api/logs/logger_provider.dart
yuzurihaaa: Question for `SignalInstanceKey`, should this be placed in `util`? Or do you have a specific place for it to be?

README.md
yuzurihaaa: Updated (manually marking this one by comment as the changes are not marked as outdated)

lib/src/api/factory/otel_api_factory.dart
yuzurihaaa: Updated (manually marking this one by comment as the changes are not marked as outdated)

lib/src/api/logs/log_record.dart
yuzurihaaa: > Instead of just names, these must have an integer value for SeverityNumber (0-24) so they can be compared. We'll need this when we integrate different loggers

afaik, with enum, this can already be compared with `Enum.key.index`

as in `SeverityNumber.unspecified.index` is 0 which correspond to the index based on https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-severitynumber

yuzurihaaa: > We need a second enum, LogLevel (don't reuse OTelLog's), that corresponds to SeverityText. TRACE, DEBUG,etc. These should have values for comparison and a method for fromSeverityNumber

1. Should I use `LogLevel.UNSPECIFIED` for `SeverityNumber.UNSPECIFIED`?
2. Does it make sense for me to use enum with argument? 

```
enum SeverityNumber {
  UNSPECIFIED(SeverityLevel.UNSPECIFIED),
  TRACE(SeverityLevel.TRACE),
  TRACE2(SeverityLevel.TRACE),
  TRACE3(SeverityLevel.TRACE),
  TRACE4(SeverityLevel.TRACE),
  DEBUG(SeverityLevel.DEBUG),
  DEBUG2(SeverityLevel.DEBUG),
  DEBUG3(SeverityLevel.DEBUG),
  DEBUG4(SeverityLevel.DEBUG),
  INFO(SeverityLevel.INFO),
  INFO2(SeverityLevel.INFO),
  INFO3(SeverityLevel.INFO),
  INFO4(SeverityLevel.INFO),
  WARN(SeverityLevel.WARN),
  WARN2(SeverityLevel.WARN),
  WARN3(SeverityLevel.WARN),
  WARN4(SeverityLevel.WARN),
  ERROR(SeverityLevel.ERROR),
  ERROR2(SeverityLevel.ERROR),
  ERROR3(SeverityLevel.ERROR),
  ERROR4(SeverityLevel.ERROR),
  FATAL(SeverityLevel.FATAL),
  FATAL2(SeverityLevel.FATAL),
  FATAL3(SeverityLevel.FATAL),
  FATAL4(SeverityLevel.FATAL);

  final SeverityLevel logLevel;

  const SeverityNumber(this.logLevel);
}

enum SeverityLevel {
  UNSPECIFIED,
  TRACE,
  DEBUG,
  INFO,
  WARN,
  ERROR,
  FATAL,
}
```

They sort of can be compared with

```
  final allValues = SeverityNumber.values.toList();
  allValues.sort((a, b) => b.logLevel.index.compareTo(a.logLevel.index));
  print(allValues);
```

yuzurihaaa: Needs elaboration on `toString` overrides. Do you mean so that if calling toString, it will show `SeverityNumber.<ENUM>` or `SeverityNumber.<ENUM>(SeverityLevel.<ENUM>)`?

michaelbushe: > afaik, with enum, this can already be compared with `Enum.key.index`
> 
> as in `SeverityNumber.unspecified.index` is 0 which correspond to the index based on https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-severitynumber

That's true but: 
- consumers may not know Dart well - they may be general o11y devs (this is called out in the spec somewhere) - so consumer won't need to know about .index and won't have to guess/confirm that we matched the order to the severity level. 
- Adding a field makes it explicit, it's self-documenting and documents the spec in the API.
- A field also self-documents code written by API consumers. (including us in the SDK and SDK users)
- Adding a field gives a handle to anyone who wants to refer to the the value as an int (and .index would not be as obvious nor as readable as .value).  
- An int allows for natural comparison instead of referring to the index.  
"Comparing Severity
In the contexts where severity participates in less-than / greater-than comparisons SeverityNumber field should be used. Special handling MAY be given to SeverityNumber=0 when it is used to represent an unspecified severity."

So.... please let's also add comparators.

toString is pretty confusing, actually.
"When an individual log record is displayed it is recommended to show both SeverityText and SeverityNumber values. A recommended combined string in this case begins with the short name followed by SeverityText in parenthesis.

For example “Informational” Syslog record will be displayed as INFO (Informational). When for a particular log record the SeverityNumber is defined but the SeverityText is missing it is recommended to only show the short name, e.g. INFO."
I think the important thing to understand is that we aren't capturing SeverityText, it will be used in the SDK and in mappers for different loggers.
Since '1' for trace means nothing, let's just always make toString() the name of the SeverityNumber.
Also, SeverityText is "also known as LogLevel" so, I was wrong, let's not use the name LogLevel in the API. But "SeverityLevel" also doesn't appear so let's go with SeverityRange.

Spec:
Comparing Severity
In the contexts where severity participates in less-than / greater-than comparisons SeverityNumber field should be used. Special handling MAY be given to SeverityNumber=0 when it is used to represent an unspecified severity.
"it is recommended to assign that severity the smallest value of the range."

So after thinking about this more...let's go with this.  I'm open to feedback, I keep looping around different solutions. 
Let's use the dart convention of lower case (well, one word camelCase) names for enums since the fields capture the output, and - unlike SpanStatus.Ok - the actual name isn't spelled out in the spec.  

I'm not sure what to do with comparison of UNSPECIFIED, what do you think?  Maybe it's ok like below, it's even lower than trace.

enum SeverityNumber {
  unspecified(0, 'UNSPECIFIED'),
  trace(1, 'TRACE1'),
  trace1(2, 'TRACE2'),
  ...
  debug(5, 'DEBUG5'),
  debug1(6, 'DEBUG6'),
   ...
   note: 'number' is not reserved but can't be used as an identifier, "SeverityNumber value" is used in the spec
  final int value;    
  final String name;
  
  const SeverityNumber(this.value, this.name);
  
  String toString() => name;
  
   @override
  int compareTo(SeverityNumber other) => value.compareTo(other.value);
  
  bool operator <(SeverityNumber other) => value < other.value;
  bool operator <=(SeverityNumber other) => value <= other.value;
  bool operator >(SeverityNumber other) => value > other.value;
  bool operator >=(SeverityNumber other) => value >= other.value;
}

enum SeverityRange {
  unspecified(SeverityNumber.unspecified),
  trace(SeverityNumber.trace), 
  debug(SeverityNumber.debug),
  ...
  fatal(SeverityNumber.fatal),
  
  LogLevel(this.name, this.severityNumber);
  
  final String name;
  //lowest number in the severity range
  final  SeverityNumber severityNumber;
  
  String toString() => name;

  int compareTo(SeverityNumber other) => severityLevel.compareTo(other.severityLevel);
  
  bool operator <(SeverityRange other) => severityNumber < other.severityNumber;
  bool operator <=(SeverityRange other) => severityNumber <= other.severityNumber;
  bool operator >(SeverityRange other) => severityNumber > other.severityNumber;
  bool operator >=(SeverityRange other) => severityNumber >= other.severityNumber;

  //these probably aren't a bad idea either
  bool operator <(SeverityNumber otherSeverityNumber) => severityNumber < other.severityNumber;
  bool operator <=(SeverityNumber otherSeverityNumber) => severityNumber <= other.severityNumber;
  bool operator >(SeverityNumber otherSeverityNumber) => severityNumber > other.severityNumber;
  bool operator >=(SeverityNumber otherSeverityNumber) => severityNumber >= other.severityNumber;

}

lib/src/api/logs/logger_provider.dart
yuzurihaaa: Ignore this. I forgot about `common`

README.md
yuzurihaaa: Added more example with somewhat closely related to real world scenario.

lib/src/api/logs/log_record.dart
```diff
@@ -0,0 +1,34 @@
...
+///
+/// Timestamp (optional)
+/// Observed Timestamp (optional)
+/// The Context associated with the LogRecord. When implicit Context is supported, then this parameter SHOULD be optional and if unspecified then MUST use current Context. When only explicit Context is supported, this parameter SHOULD be required.
+/// Severity Number (optional)
+/// Severity Text (optional)
+/// Body (optional)
+/// Attributes (optional)
+/// Event Name (optional)
+abstract class LogRecord {
```
yuzurihaaa, line 19: info: Realised that this is under API section https://opentelemetry.io/docs/specs/otel/logs/api/#emit-a-logrecord

lib/src/api/logs/severity.dart
```diff
@@ -0,0 +1,60 @@
+enum Severity {
```
yuzurihaaa, line 1: info: I am also using https://github.com/open-telemetry/opentelemetry-java/blob/9f4f40532fded48f92456db25ca561292215cf14/api/all/src/main/java/io/opentelemetry/api/logs/Severity.java as reference for this one.

README.md
michaelbushe: These are really nice.  Thank you.

lib/src/api/logs/logger_provider.dart
michaelbushe: Better now but you added these:
/// final loggerProvider = OTel.loggerProvider();
/// final logger = loggerProvider.getLogger('');

Should be (+API + log name)
/// final loggerProvider = OTelAPI.loggerProvider();
/// final logger = loggerProvider.getLogger('my-service-logger');

Empty log names throw errors, per spec.  
final validatedName = name.isEmpty ? '' : name;
    if (validatedName.isEmpty) {
      OTelLog.warn(
          'Invalid log name provided; using empty string as fallback.');
    }
    
   Examples are important. They are copied by humans and LLMs.

lib/src/api/logs/logger_provider.dart
michaelbushe: Nice.  I am going to make a push to your branch with some minor changes, including using SignalInstanceKey in traces and metrics.  This required adding a new Signal enum to differentiate one from another (even though it's  unlikely there will ever be a set with signal instances of different types).

lib/src/api/logs/severity.dart
michaelbushe: I love Java. Dart is a better Java (both lang specs written by Gilead Bracha).  
We can do better: Dart has operators, Dart enums can have hashCode/Equals.
This Java class looks is missing a toString(), IMHO.

lib/src/api/logs/log_record.dart
michaelbushe: Info back: 
https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-timestamp
Field: Timestamp
Type: Timestamp, uint64 nanoseconds since Unix epoch.
Description: Time when the event occurred measured by the origin clock, i.e. the time at the source. This field is optional, it may be missing if the source timestamp is unknown.
----
We probably should use nanosecond time. I'm looking at using time_machine Instant instead of DateTime but time_machine not at 1.0 (and neither are we).  I'll need to check other uses of DateTime in traces and metrics to see we have to make breaking changes.

michaelbushe:
We're getting close. Thank you!

lib/src/api/logs/log_record.dart
michaelbushe: After looking more closely, the correct type for typestamps is Int64.
import 'package:fixnum/fixnum.dart';

Int64? get timestamp;
Int64? get observedTimestamp;

This matches the protobufs in the SDK:
https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/blob/30b936c354ae31f966d60ae74052bbf74ed1ff23/lib/proto/logs/v1/logs.pb.dart#L294

Please also add a time.dart in ./util with: 
// Helper function for getting current time in OTel nano format with microsecond precision
Int64 nowAsNanos() {
  return Int64(DateTime.now().microsecondsSinceEpoch * 1000);
}

// Extension for getting current time in OTel nano format with microsecond precision
extension DateTimeToNanos on DateTime {
  // returns the current time in OTel nano format with microsecond precision
  Int64 toNanos() {
    return Int64(microsecondsSinceEpoch * 1000);
  }
}

lib/src/api/logs/logger_provider.dart
yuzurihaaa: > Empty log names throw errors, per spec.


From [here](https://opentelemetry.io/docs/specs/otel/logs/sdk/#logger-creation), 

> In the case where an invalid name (null or empty string) is specified, a working Logger MUST be returned as a fallback rather than returning null or throwing an exception, its name SHOULD keep the original invalid value, and a message reporting that the specified value is invalid SHOULD be logged.

cmiiw, I don't think I should throw error here 🤔

lib/src/api/logs/log_record.dart
yuzurihaaa: Updated.

yuzurihaaa:
Updated some based on review. Still getting the unit test to pass at certain percentage threshold.

yuzurihaaa:
Added unit test... generated from copilot 😅

michaelbushe:
I'll get to this review later this week. Thank you!

michaelbushe (approved):
Looks great! Thank you for your work.  Apologies for the recent review delay.  This is good to go.  Thanks again!

PR #5, opened 2026-03-21
Move to and fix much more strict lints

kevmoo:
I ignored a few annoying ones


michaelbushe (approved):
Thank you @kevmoo, it's great to aligned with the dart team lints.

PR #6, opened 2026-03-21
Fix web test failures for dart2wasm and dart2js

kevmoo:
- Update conditional imports to use dart.library.js_interop for wasm.
- Use Int64 for nanosecond timestamps to prevent precision loss on JS.
- Make time-based matchers and tests inclusive to handle low-resolution timers.

### DIFF
--- a/lib/src/api/context/context.dart
+++ b/lib/src/api/context/context.dart
@@ -13,7 +13,8 @@ import '../trace/span.dart';
 import '../trace/span_context.dart';
 import 'context_key.dart';
 // Use conditional imports to handle platform differences
-import 'isolate_support.dart' if (dart.library.js) 'isolate_support_web.dart';
+import 'isolate_support.dart'
+    if (dart.library.js_interop) 'isolate_support_web.dart';
 
 part 'context_create.dart';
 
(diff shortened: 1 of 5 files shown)


kevmoo:
I'd recommend enabling tests for web bits on CI: `dart test -p chrome -c dart2wasm,dart2js`

lib/src/api/context/context.dart
```diff
@@ -13,7 +13,8 @@ import '../trace/span.dart';
 import '../trace/span_context.dart';
 import 'context_key.dart';
 // Use conditional imports to handle platform differences
-import 'isolate_support.dart' if (dart.library.js) 'isolate_support_web.dart';
```
kevmoo, line 16: `dart.library.js` only works with JavaScript, not WebAssembly.

michaelbushe: Good catch! I upgraded in the SDK but left this dangling in the API. Thank you.

michaelbushe (approved):
Thanks @kevmoo. I first merge #9 to expose the test failures.  This PR fixes all of them.

Notes for me to follow up on: 
Timestamp.now/fromDateTime/toDateTime becoming Int64 is a breaking API change for downstream consumers (the SDK package will need a small update); will note in CHANGELOG.

Also noticed that nowAsNanos() in [lib/src/util/time.dart](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/blob/main/lib/src/util/time.dart) does Int64(... * 1000) instead of Int64(...) * 1000 — same JS-precision pattern, not touched by this PR. Tracking as a follow-up.

Approving and merging.

kevmoo:
🙏

PR #7, opened 2026-03-21
Refactor context management and trace activation to follow OTel spec and improve async reliability

kevmoo:
- Breaking: tracer.startSpan() no longer automatically activates the span in the current context, aligning with the OpenTelemetry specification.
- Breaking: Deprecated the static Context.current setter in favor of Context.run() and Context.runSync() to ensure correct asynchronous propagation via Zones.
- Refactored APITracer.withSpan() and withSpanAsync() to use Zone-based context propagation (runSync/run).
- Added Context.runSync() for activating contexts in synchronous code blocks.
- Fixed a bug in APITracer.createSpan() where parent spans from the provided or current context were ignored.
- Optimized Context.serialize() by adding isTransferable to ContextKey to skip unnecessary serializability checks for non-transferable keys.
- Fixed Context.runIsolate() to correctly serialize the specific context instance it was called on instead of the global current context.
- Updated entire test suite to reflect modern context patterns and added isolate_transfer_test.dart.
- Updated README and examples to demonstrate modern Zone-based context management.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -80,4 +80,4 @@ jobs:
       - name: Install dependencies
         run: dart pub get
       - name: Run web tests (dart2js + dart2wasm)
-        run: dart test -p chrome -c dart2js,dart2wasm --retry=2
+        run: dart test -p chrome -c dart2js,dart2wasm
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,17 +5,27 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [1.0.0-wip] - TBD
+## [1.0.0-alpha.1] - TBD
+
+### Added
+- `Context.run()` and `Context.runSync()` — Zone-based implicit context propagation. These are the spec-aligned way to attach a context for a scope of execution and ensure it propagates correctly across `await`s and async callbacks.
+- `isTransferable` flag on `ContextKey` (default `false`) to opt custom keys into cross-isolate transfer via `Context.runIsolate()`, default is false.
 
 ### Changed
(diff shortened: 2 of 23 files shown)


kevmoo:
Based off https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/5

michaelbushe:
Hey @kevmoo — picked up the rebase, format, and a few small fixups (CHANGELOG, the nowAsNanos precision fix, currentWithBaggage purification, runIsolate cleanup). Marking ready for review and squash-merging shortly. Thanks for the substantive work on the Zone-based propagation — it's the right direction for the package and fixes some goofiness. 🙏. Note that this is the API so APISpan is rarely if ever used.  I'll port these changes to the SDK after the alpha release.

michaelbushe (approved):
Thank you @kevmoo — very grateful. The Zone-based context propagation is the right architectural move for the package, matching the OTel spec's implicit propagation API and how the Java/Python/JS SDKs handle it.

I picked up a few small fixups on top of your commit:
Reverted _getAndCacheOTelFactory rename back to "OTel" casing for consistency with the rest of the codebase (OTelAPI, APISpan, etc.) and the OpenTelemetry API.  Dart and OTel have style tension and I went with OTel's style across the API and SDKs.  

Fixed nowAsNanos() precision bug on JS in [lib/src/util/time.dart](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/blob/try-rebase/lib/src/util/time.dart) — Int64 wrap now happens before the * 1000 multiplication, same pattern as your Timestamp.fromDateTime fix in [#6](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/issues/6).
Made Context.currentWithBaggage() pure (no longer mutates Context.current). Fits the new Zone-based model. All seven SDK example calls only rely on the non-null Baggage return, so they're unaffected.
Cleaned up Context.runIsolate(): dropped the deprecated Context.current = setter call (redundant with the next line), and removed the try/finally that could leak Zone state into the parent isolate's static field.
Removed the leftover // ... rest of logger calls placeholder in the example.
Expanded CHANGELOG with the breaking changes called out.
SDK update for these breakages is next. Approving and squash-merging.  Thanks again. 💙

PR #9, opened 2026-05-05
ci: fixed retry attempt

michaelbushe:
retry needs to be in yaml for web platform, will follow up.  The correct failure will show now in ci and will be fixed by #6.


michaelbushe:
Merging the expected failures of 4 tests that #6 should turn green, shortly.

PR #11, opened 2026-05-07
fix: trace continuity across Context.runIsolate()

michaelbushe:
Summary

A SpanContext that crosses an isolate boundary via Context.runIsolate() should be isRemote=true on the receiving side, matching the OpenTelemetry specification: "Each propagators' deserialization must set IsRemote to true on a parent SpanContext" (trace/api.md
  §SpanContext (https://opentelemetry.io/docs/specs/otel/trace/api/#spancontext)). Previously the parent isolate's local SpanContext (with isRemote=false) was restored verbatim, so any code that uses the OTel-spec convention of treating remote SpanContexts as parents (e.g. APITracer.createSpan's remote-context branch) skipped the propagated SpanContext and produced a fresh root span.

Fix
In Context.runIsolate, after deserialize(serializedContext), if the resulting SpanContext has isRemote=false, rebuild it with isRemote=true before attaching to the Zone and statics. Context.deserialize itself is left round-trip-faithful — only the cross-isolate path overrides remoteness, since that's the only path where remoteness is unambiguous.                               

Added (initially failing) test for Context Isolate Transfer › SpanContext is marked isRemote=true after crossing isolate boundary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.0.0-beta.1] - 2026-05-07
+
+### Fixed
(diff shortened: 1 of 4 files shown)


michaelbushe:
@kevmoo Tagging you because this patches your isolate code from #7 —  
  non-blocking.

PR #17, opened 2026-05-10
chore(release): auto-checkout tag + auto-publish in release.dart

michaelbushe:
Closes a footgun in the previous release flow: the script left HEAD on the next-wip bump commit, so a follow-up `dart pub publish` saw the `-wip` pubspec and offered to publish the wip version. (Easy to miss the version in the prompt — pre-release identifiers in semver versions look like normal version suffixes.)

Now after the bump commit, release.dart:
  - records the current branch
  - checks out the v$release tag (so pub publish reads the right tree)
  - runs `dart pub publish` with stdio inherited so pub.dev's own Y/N confirmation reaches the user's terminal
  - returns the working tree to the original branch

New flags: `--no-publish` to stop after the local commits, `--skip-tests` to skip `dart test` (e.g. when CI runs it separately). Existing `--next` and `--yes` unchanged.

CONTRIBUTING.md and PUBLICATION_CHECKLIST.md updated to reflect that the script now publishes by default.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -154,19 +154,26 @@ and CHANGELOG entries during development land under
 ```bash
 dart tool/release.dart                       # auto-bump trailing number
 dart tool/release.dart --next 1.2.0-beta     # override the next dev version
-dart tool/release.dart --yes                 # non-interactive (CI)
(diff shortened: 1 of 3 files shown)


michaelbushe:
Superseded by #16, which already merged the same
  change.

PR #18, opened 2026-05-10
chore(release): clarify that pub.dev still prompts before uploading

michaelbushe:
## Summary

The pre-release summary printed:

```
  Publish:   yes (auto)
```

which reads as "non-interactive / no further prompts". `dart pub publish` actually still prompts for a final y/N before uploading — the "auto" suffix only meant "the script auto-invokes pub publish for you". User-confusing.

After:

```
  Publish:   yes — pub.dev will prompt before uploading
```

(or `no — local commits + tag only` when `--no-publish` is set)

Plus the script's own gate now reads `Continue with the local release steps? [y/N]` so the y is clearly scoped to "make the local commits + tag", not "publish to pub.dev". And a one-line reminder is printed before invoking pub publish so users aren't surprised when pub.dev's own confirmation appears.

### DIFF
--- a/tool/release.dart
+++ b/tool/release.dart
@@ -100,10 +100,17 @@ Future<void> main(List<String> args) async {
     );
     _runOrThrow('dart', ['pub', 'get'], silent: true);
     _runOrThrow('dart', ['analyze']);
-    if (!flags.skipTests) {
-      _runOrThrow('dart', ['test']);
+    if (flags.skipTests) {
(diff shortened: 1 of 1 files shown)


michaelbushe:
Pushed an additional commit (`1a2cf57`) mirroring the SDK's `tool/test.sh` detection. No-op for the API itself since this repo doesn't ship a `tool/test.sh`, but keeps the two release scripts byte-for-byte identical.

PR #26, opened 2026-05-18
doc: initialize new behavior, CHANGELOG

michaelbushe:
Documenting the new initialization behavior


lib/src/api/otel_api.dart
```diff
@@ -86,7 +102,7 @@ class OTelAPI {
       String? serviceName = OTelAPI.defaultServiceName,
       String? serviceVersion = OTelAPI.defaultServiceVersion,
       OTelFactoryCreationFunction? oTelFactoryCreationFunction}) {
-    if (OTelFactory.otelFactory != null) {
+    if (_userInitialized) {
```
benjaben, line 105: This is a bit cleaner and more in line with how other SDKs work 👍

benjaben (approved):

michaelbushe:
This is not a good solution.

PR #29, opened 2026-07-08
fix: Context getters no longer throw before OTelAPI.initialize()

### LINKED ISSUE #28: [Spec Violation] Context getters throw StateError when OpenTelemetry is uninitialized
Per the [OpenTelemetry Library Guidelines](https://opentelemetry.io/docs/specs/otel/library-guidelines/):
> *"The OpenTelemetry API MUST NOT require explicit initialization... When an SDK is not installed or initialized, API calls MUST NOT fail or throw exceptions; they MUST operate as no-ops."*

In `lib/src/api/context/context.dart`, accessing `Context.root` or `Context.current` prior to initialization executes:
```dart
static OTelFactory _getAndCacheOTelFactory() {
  if (_otelFactory != null) return _otelFactory!;
  if (OTelFactory.otelFactory == null) {
    throw StateError('Call initialize() first.'); // <-- SPEC VIOLATION
  }
  return _otelFactory = OTelFactory.otelFactory!;
}
```

If any third-party library inspects `Context.current` during module load or before the application explicitly calls `initialize()`, the entire Dart VM crashes with unhandled `StateError`s.

**Expected Behavior:**  
Accessing context getters prior to SDK initialization must gracefully return the default No-Op context without throwing exceptions (for instance, by delegating `Context._getAndCacheOTelFactory()` to `OTelAPI._getAndCacheOtelFactory()`).

benjaben:
## Summary
- `Context.root`/`Context.current` (and anything that touches them, e.g. `withBaggage`) threw `StateError('Call initialize() first.')` when accessed before an SDK/API factory was installed, violating the [OTel spec requirement](https://opentelemetry.io/docs/specs/otel/library-guidelines/) that the API operate as a no-op without explicit initialization.
- Added `OTelFactory.getOrCreateDefault()` as the single choke point for lazily installing the No-Op API factory, and pointed `Context._getAndCacheOTelFactory()` at it.
- `OTelAPI._getAndCacheOtelFactory()` already had equivalent self-init logic; it now delegates to the same shared helper instead of duplicating it.

Fixes #28.

## Test plan
- [x] `dart analyze lib/ test/` — no issues
- [x] `dart test` — all tests pass
- [x] New regression test (`test/unit/api/context/context_uninitialized_test.dart`) added in its own file so it runs before any other test calls `OTelAPI.initialize()`; confirmed it reproduces the original `StateError` when the fix is reverted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-beta.8-wip]
 
+### Fixed
+- `Context.root` / `Context.current` (and other Context APIs) threw
+  `StateError('Call initialize() first.')` when accessed before
+  `OTelAPI.initialize()`, violating the OTel spec requirement that the API
+  operate as a no-op without an SDK installed. They now lazily install the
+  No-Op API factory, matching `OTelAPI`'s existing behavior.
+
 ## [1.0.0-beta.7] - 2026-05-18
 
 ## [1.0.0-beta.6] - 2026-05-11
--- a/lib/src/api/context/context.dart
+++ b/lib/src/api/context/context.dart
@@ -40,16 +40,12 @@ class Context {
 
   /// Gets and caches the OTelFactory instance.
   ///
-  /// Retrieves the global OTelFactory instance and caches it for future use.
-  /// Throws a StateError if OpenTelemetry has not been initialized.
+  /// Retrieves the global OTelFactory instance, lazily installing the No-Op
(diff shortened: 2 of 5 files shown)


michaelbushe:
Thanks @benjaben! GitHub doesn't honor maintainer-edit pushes on org-owned forks, so we couldn't update this branch directly — your commit landed, authorship intact, in #31, rebased onto main with two follow-ups that keep the lazily-installed no-op factory upgradeable (Context re-reads the global factory on every access, and OTelAPI.initialize() replaces an isAPIFactory factory instead of throwing). Closing in favor of #31; this ships in 1.0.0-beta.8. Much appreciated!

PR #31, opened 2026-07-11
fix: Context works as a no-op before initialization and stays upgradeable

### LINKED ISSUE #28: [Spec Violation] Context getters throw StateError when OpenTelemetry is uninitialized

michaelbushe:
## Summary

Carries #29 by @benjaben (rebased onto main, his commit and authorship intact — pushing to the org-fork branch isn't permitted, so this PR supersedes it) plus two follow-up fixes that keep the lazily-installed no-op factory upgradeable.

## Changes

**From #29 (@benjaben):** `Context.root` / `Context.current` threw `StateError('Call initialize() first.')` when accessed before `OTelAPI.initialize()`, violating the OTel spec requirement that the API operate as a no-op without an SDK. They now lazily install the no-op API factory via a single choke point, `OTelFactory.getOrCreateDefault()`, shared with `OTelAPI`. Fixes #28.

**Follow-ups (this PR):**
- `Context` re-reads `OTelFactory.otelFactory` on every access (matching `OTelAPI`) instead of keeping the first factory it saw — otherwise a no-op cached before SDK initialization would shadow the SDK's factory forever, the same staleness class as dartastic_opentelemetry#50.
- `OTelAPI.initialize()` replaces an installed factory when `isAPIFactory` is true instead of throwing — merely touching `Context.current` before initializing no longer permanently blocks explicit initialization. Only a real (non-API) factory still triggers the initialize-once `StateError`. **Behavior change:** re-initializing over a no-op API factory replaces it and applies the new configuration.

Both follow-up defects are pinned by regression tests that were committed failing first (`test: prove lazy no-op install must stay upgradeable`) and turned green by the fix commit:
- `context_then_initialize_test.dart` — pre-init `Context.current` access, then `OTelAPI.initialize()` succeeds.
- `context_factory_upgrade_test.dart` — after a real factory replaces the no-op, `Context` delegates to it (with a built-in control showing `OTelAPI` already did).

## Validation

- `dart analyze` — clean
- `dart format --set-exit-if-changed .` — clean
- `dart test` — 735 tests passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   factory installed when API code runs first, instead of relying on
   `runtimeType` checks (see dartastic_opentelemetry #50 / PR #53).
 
+### Fixed
+- `Context.root` / `Context.current` (and other Context APIs) threw
+  `StateError('Call initialize() first.')` when accessed before
+  `OTelAPI.initialize()`, violating the OTel spec requirement that the API
+  operate as a no-op without an SDK installed. They now lazily install the
+  No-Op API factory, matching `OTelAPI`'s existing behavior.
+
 ## [1.0.0-beta.7] - 2026-05-18
 
 ## [1.0.0-beta.6] - 2026-05-11
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `StateError('Call initialize() first.')` when accessed before
   `OTelAPI.initialize()`, violating the OTel spec requirement that the API
   operate as a no-op without an SDK installed. They now lazily install the
-  No-Op API factory, matching `OTelAPI`'s existing behavior.
+  No-Op API factory, matching `OTelAPI`'s existing behavior. Thanks
+  @benjaben.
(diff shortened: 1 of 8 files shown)


michaelbushe:
Verified against the 1.1.0-beta.7-wip, 1793 SDK tests passed against the new API.

PR #36, opened 2026-07-11
fix: Baggage spec compliance — accept empty values, never throw

michaelbushe:
## Summary

Documents and fixes a spec violation found auditing against `specification/baggage/api.md` and `error-handling.md`. TDD: first commit pins the spec contracts red (8 failing), second turns them green.

## The violation

- "Baggage **values** are any valid UTF-8 strings. Language API MUST accept any valid UTF-8 string as baggage **value** in `Set` and return the same value from `Get`." — the empty string is a valid UTF-8 string, but `Baggage` threw `ArgumentError` on it, and both `fromJson` paths silently dropped empty values.
- "API methods MUST NOT throw unhandled exceptions when used incorrectly by end users" (error-handling.md) — empty names threw `ArgumentError`; now ignored with a warning (names must be non-empty per spec, so the entry is invalid — but rejection is a log, not a crash).
- "The Baggage API MUST be fully functional in the absence of an installed SDK." — `copyWith`/`copyWithout`/`copyWithBaggage` threw `StateError('Call initialize() first.')` after reset; they now lazily install the no-op factory like the rest of the API.

## Merge order

Stacked on #35 (which stacks on #34): merge #34 → #35 → this.

## Validation

- `dart test` — 755 tests passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -334,6 +334,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the API minted random valid IDs and returned recording spans (#40).
   SDK span creation is unaffected: the no-op behavior applies only when
   the installed factory `isAPIFactory`.
+- **`Baggage` now follows the spec for values, names, and no-SDK use.**
+  Per the Baggage API spec, values are any valid UTF-8 string — the empty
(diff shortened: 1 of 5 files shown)


michaelbushe:

PR #37, opened 2026-07-11
fix: TraceState spec compliance — ignore invalid input, never throw

michaelbushe:
## Summary

Documents and fixes a spec violation found auditing against `specification/trace/api.md`. TDD: first commit pins the spec contracts red (5 failing), second turns them green.

## The violation

- "Every mutating operations MUST validate input parameters. If invalid value is passed the operation MUST NOT return `TraceState` containing invalid data and MUST follow the general error handling guidelines" — and error-handling.md: "API methods MUST NOT throw unhandled exceptions when used incorrectly by end users." `TraceState.put` threw `ArgumentError` on invalid keys/values; it now warns and returns the state unchanged, so the state stays valid and the caller never crashes.
- `put`/`remove` also threw `StateError('Call initialize() first.')` before initialization; they now lazily install the no-op factory like the rest of the API.

## Merge order

Stacked on #36: merge #36 first, then this. Note: `fix/tracestate-tenant-key-w3c` (in-flight from another session) touches the same file — whichever merges second will need a small rebase.

## Validation

- `dart test` — 760 tests passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -342,6 +342,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `copyWithBaggage` work without an installed SDK (previously `StateError`),
   per "The Baggage API MUST be fully functional in the absence of an
   installed SDK."
+- **`TraceState.put` / `remove` never throw.** Per the trace API spec,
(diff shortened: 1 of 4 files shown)


michaelbushe:

PR #38, opened 2026-07-11
fix: validate tracestate tenant/system keys per W3C grammar

michaelbushe:
## Summary

The multi-tenant `tracestate` key branch of `TraceState._isValidKey` diverged from the [W3C Trace Context key grammar](https://www.w3.org/TR/trace-context/#key) in two ways:

- **tenant-id** was required to start with a lowercase letter, but W3C allows a leading digit (`tenant-id = ( lcalpha / DIGIT ) ...`) — so a valid key like `9legacy@acme` was wrongly rejected.
- **No length bounds** — W3C caps `tenant-id` at 241 chars and `system-id` at 14; over-long ids were silently accepted.

The fix replaces the shared per-part regex with distinct `_tenantIdFormat` / `_systemIdFormat` patterns, and also caps the simple-key regex at 256 chars to match the spec.

## Validation

- 4 red tests pin the contracts (committed first, failing against the old regex); the fix turns them green.
- `dart analyze` — clean
- `dart format` — clean
- `dart test` — 753 passing (was 747)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   longer bypass the factory via the static `attrsFromMap` "cheat" (obsolete
   since the beta.8 lazy-install lifecycle); a factory that overrides
   `attributesFromMap` is now respected on all three paths (#33).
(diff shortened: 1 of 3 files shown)


michaelbushe:

PR #39, opened 2026-07-11
fix: provider accessors use spec-mandated safe defaults, never throw

michaelbushe:
## Summary

Documents and fixes a spec violation found auditing against `specification/trace/api.md` and `error-handling.md`. TDD: first commit pins the spec contracts red (6 failing), second turns them green.

## The violation

- trace/api.md, Get a Tracer: "In case an invalid name (null or empty string) is specified, a working Tracer implementation MUST be returned as a fallback rather than returning null or throwing an exception" — `OTelAPI.tracerProvider('')` / `meterProvider('')` / `loggerProvider('')` threw `ArgumentError`; they now warn and return the global default (exactly what the old error message told users to do manually).
- error-handling.md: "API methods MUST NOT throw unhandled exceptions when used incorrectly by end users" — `getTracer` / `getMeter` / `getLogger` threw `StateError` after provider shutdown; they now warn and return a no-op instance.

## Merge order

Stacked on #37 (→ #36): merge #36 → #37 → this.

## Validation

- `dart test` — 766 tests passing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -348,6 +348,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   invalid keys/values are now ignored with a warning (previously
   `ArgumentError`), and both operations work without an installed SDK
   (previously `StateError`).
+- **Provider accessors use safe defaults instead of throwing.** Per the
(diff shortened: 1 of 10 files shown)


michaelbushe:

Issue #40, opened 2026-07-11
No-SDK trace behavior violates trace/api.md: random IDs, recording spans, no SpanContext wrapper


michaelbushe:
Found auditing against `specification/trace/api.md`, "Behavior of the API in the absence of an installed SDK" and "Wrapping a SpanContext in a Span".

## The spec

- "The API MUST return a non-recording `Span` with the `SpanContext` in the parent `Context` (whether explicitly given or implicit current)."
- "If the parent `Context` contains no `Span`, an empty non-recording Span MUST be returned instead (i.e., having a `SpanContext` with all-zero Span and Trace IDs, empty Tracestate, and unsampled TraceFlags)."
- "The API MUST provide an operation for wrapping a `SpanContext` with an object implementing the `Span` interface" — `GetContext` returns the wrapped context, `IsRecording` returns false, everything else no-op; "SHOULD be named `NonRecordingSpan`."

## The violations (API-only, no SDK installed)

- `APITracer.createSpan` root path generates **random** trace/span IDs (`lib/src/api/trace/tracer.dart` ~180-186) instead of all-zero IDs.
- With a parent in context, it mints a **new span ID** for the child instead of returning a span carrying the parent's `SpanContext` unchanged.
- `APISpan.isRecording => !isEnded` (`lib/src/api/trace/span.dart` ~155), so no-SDK spans report `isRecording == true` while active; the spec requires non-recording.
- No `NonRecordingSpan` / SpanContext-wrapping operation exists.

## Design constraint for the fix

The SDK's `Tracer` **delegates `createSpan`/`startSpan` to `APITracer`** (`dartastic_opentelemetry/lib/src/trace/tracer.dart` ~200, ~368), using it as the ID-generation engine for real spans. The no-op behavior therefore can't be unconditional — it must apply only when the installed factory `isAPIFactory` (the beta.8 getter), or the delegation needs restructuring (e.g., a dedicated no-op tracer, with ID generation moved SDK-side). Either direction changes many existing API-repo tests that assert valid random IDs pre-SDK.

Skipped compliance tests pinning the required behavior: `test/unit/api/trace/no_sdk_span_behavior_test.dart` (see the PR referencing this issue; remove the `skip:`s to reproduce).

PR #41, opened 2026-07-11
test: document no-SDK span behavior spec violations (#40)

michaelbushe:
## Summary

Documents (does not fix) the spec violations in #40, found auditing against `specification/trace/api.md`, "Behavior of the API in the absence of an installed SDK". TDD: the first commit pins the contracts red (3 failing, with the observed wrong values in the test output), the second skips them pointing at #40 so CI stays green until the fix lands.

## The violations (API-only, no SDK)

- Parentless spans get random valid trace/span IDs — the spec requires "a `SpanContext` with all-zero Span and Trace IDs, empty Tracestate, and unsampled TraceFlags".
- Spans report `isRecording == true` — the spec requires non-recording spans.
- A span with a parent gets a freshly minted span ID — the spec requires "a non-recording `Span` with the `SpanContext` in the parent `Context`" unchanged.

## Why not fixed here

The SDK's `Tracer` delegates `createSpan`/`startSpan` to `APITracer` as its ID-generation engine, so making the API no-op unconditionally would break real SDK spans. The fix needs a design decision (condition on `isAPIFactory`, or restructure the delegation and add `NonRecordingSpan`) — see #40. Removing the `skip:` in `test/unit/api/trace/no_sdk_span_behavior_test.dart` reproduces all three.

## Validation

- `dart analyze` / `dart test` — clean (3 skipped)

### DIFF
--- a/test/unit/api/trace/no_sdk_span_behavior_test.dart
+++ b/test/unit/api/trace/no_sdk_span_behavior_test.dart
@@ -1,0 +1,56 @@
+// Licensed under the Apache License, Version 2.0
+
+import 'package:dartastic_opentelemetry_api/dartastic_opentelemetry_api.dart';
+import 'package:test/test.dart';
+
+// Spec-compliance tests for trace/api.md, "Behavior of the API in the
+// absence of an installed SDK":
+//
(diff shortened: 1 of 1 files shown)


michaelbushe:

Issue #42, opened 2026-07-11
No global propagator API — context/api-propagators.md requires get/set with no-op default


michaelbushe:
Found auditing against `specification/context/api-propagators.md`, "Global Propagators".

## The spec

- "The OpenTelemetry API MUST provide a way to obtain a propagator for each supported `Propagator` type" — with a Get/Set global "method MUST exist for each supported `Propagator` type."
- "The OpenTelemetry API MUST use no-op propagators unless explicitly configured otherwise."

## The violation

Neither the API package nor the SDK exposes a global `TextMapPropagator` accessor, and there is no no-op `TextMapPropagator` implementation. The interface exists (`lib/src/api/context/propagation/text_map_propagator.dart`, generic `TextMapPropagator<C, V>`) and the SDK ships W3C trace-context and baggage propagators, but instrumentation libraries have no spec-mandated way to obtain "the" propagator — they must be handed one explicitly.

## Design notes for the fix

- A global holder over the generic `TextMapPropagator<C, V>` is type-lossy (`TextMapPropagator<dynamic, dynamic>`); worth deciding whether the global surface should be non-generic (as in other OTel languages) before implementing.
- Needs a no-op `TextMapPropagator` (extract returns the passed context unchanged; inject writes nothing) as the default.
- SDK `initialize()` would set the global (and eventually honor `OTEL_PROPAGATORS`).

Skipped compliance test: `test/unit/api/context/propagation/global_propagator_test.dart` (see the PR referencing this issue).

PR #43, opened 2026-07-11
test: document missing global propagator API (#42)

michaelbushe:
## Summary

Documents (does not fix) the spec violation in #42, found auditing against `specification/context/api-propagators.md`, "Global Propagators". TDD: first commit pins the requirement red, second skips it pointing at #42 so CI stays green until the design lands.

## The violation

- "The OpenTelemetry API MUST provide a way to obtain a propagator for each supported `Propagator` type" with global Get/Set, and "MUST use no-op propagators unless explicitly configured otherwise" — neither a global accessor nor a no-op `TextMapPropagator` exists; instrumentation libraries cannot obtain "the" propagator.

## Why not fixed here

The existing `TextMapPropagator<C, V>` is generic, so a global holder is type-lossy; whether the global surface should be non-generic (as in other OTel languages) is a design decision — see #42.

## Validation

- `dart analyze` / `dart test` — clean (1 skipped)

### DIFF
--- a/test/unit/api/context/propagation/global_propagator_test.dart
+++ b/test/unit/api/context/propagation/global_propagator_test.dart
@@ -1,4 +1,5 @@
-// Licensed under the Apache License, Version 2.0
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
 
 import 'package:test/test.dart';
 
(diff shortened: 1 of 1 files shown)


michaelbushe:

harshitt13 (approved):

PR #47, opened 2026-07-11
chore: enforce *Create classes as package-internal

michaelbushe:
Companion to MindfulSoftwareLLC/dartastic_opentelemetry#62 (which applied the same lockdown to the SDK's Create classes).

## Why

The `*Create` classes were internal by doc-comment convention only. They're public classes in `part` files of barrel-exported libraries, so `AttributesCreate.create()` etc. were callable by any consumer — an open factory-bypass door, contradicting the beta.8/9 removal of factory cheat paths. (The SDK itself used `AttributesCreate` that way until #62 removed it.) The `part of` + private-constructor pattern protects the *constructors*; the public Create wrappers were exactly a public door to them.

## What

- `@internal` (package:meta) on all 28 `*Create` classes — analyzer rejects outside-package use (`invalid_use_of_internal_member`), dartdoc hides them
- Barrel `hide`s them, enforced by `invalid_export_of_internal_element`
- `meta` imports added to host libraries
- Factories, same-package code, and tests are unaffected

## Validation

- `dart analyze lib test`: no issues
- Full suite: 753 passed, 0 failures
- The SDK no longer references any API Create class (verified in #62), so releasing this as beta.10 breaks nothing downstream

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **`*Create` classes are now `@internal` and hidden from the public library.**
+  They were internal by doc-comment convention only; the barrel exported them,
+  so `AttributesCreate.create()` etc. were callable by any consumer — an open
+  factory-bypass door, contradicting the beta.8/9 removal of factory cheat
+  paths. The analyzer now rejects outside-package use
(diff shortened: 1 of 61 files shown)


michaelbushe:
@kevmoo If you have time  please take a look at my use of Dart Language features to enforce creation of API (and SDK) objects via `OTelFactory`.

PR #103, opened 2026-08-24
fix(api): store empty string and empty array attribute values per OTel spec

### LINKED ISSUE #80: [BUG] Common: empty string and empty array attribute values are rejected and throw
> **Migrated from MindfulSoftwareLLC/dartastic_opentelemetry#233** — the fix belongs entirely in `dartastic_opentelemetry_api`; the SDK issue is closed as moved.

## Spec requirement

**Requirement level:** MUST / MUST NOT

> AnyValues expressing an empty value, a numerical value of zero, an empty string,
> or an empty array are considered meaningful and MUST be stored and passed on to
> processors / exporters.

The error handling rules add:

>
> 1. API methods MUST NOT throw unhandled exceptions when used incorrectly by end users.

## The problem

This issue covers two spec clauses that share one constructor and one fix.

**The value must be stored.** An empty string and an empty array are meaningful values. This code treats both as invalid.

**The code must not throw.** The `Attribute` constructor throws an `ArgumentError` on either value. Public entry points reach that constructor, so `OTel.attributeString('key', '')` produces an unhandled exception during ordinary instrumentation.

Three more paths drop these values in silence. `attrsFromMap` skips an empty string and an empty list. `Attributes.fromJson` drops an empty list and states in a log line that the spec forbids them, which is the opposite of the rule. A doc comment repeats the wrong behavior.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/common/attribute.dart:23`

```dart
Attribute._(String key, T value)
    : _key = key,
      _value = value {
  if (_value is String && (_value as String).isEmpty) {
    throw ArgumentError('Attribute _value must not be an empty string');
  }
  if (_value is List) {
    final valueList = _value as List;
    if (valueList.isEmpty) {
      throw ArgumentError('Attribute _value list must not be empty');
    }
    ...
```

The silent drops are at `factory/otel_api_factory.dart:173` and `:192`, and at `common/attributes.dart:79-81`. The wrong doc comment is at `common/attributes.dart:327`.

## How to correct it

1. Remove both throws from the constructor.
2. Store an empty string and an empty list.
3. Remove the skip in `attrsFromMap` and in `fromJson`.
4. Correct the log message and the doc comment.

NOTE: This is not a fail-fast case at startup, which the spec does permit. Attribute creation runs during normal instrumentation.

## Notes

- Source finding files: `fable-ultra/common-01-attribute-ctor-throws.md`, `fable-ultra/common-02-empty-values-dropped.md`, `fable-medium/common-empty-values-rejected.md`, `grok/common/001-empty-string-attribute-rejected.md`, `grok/common/002-empty-list-attribute-rejected.md`, `grok/common/003-attribute-constructor-throws.md`
- Parent issue: #95

abidiahmedcom:
## What

Fixes #80. Empty string and empty array attribute values are now stored instead of throwing or being silently dropped.

## Why

The OTel specification ([Common Specification, v1.60.0](https://opentelemetry.io/docs/specs/otel/common/#attribute)) requires:

> AnyValues expressing an empty value, a numerical value of zero, an empty string, or an empty array are considered meaningful and MUST be stored and passed on to processors / exporters.

Previously:
- `OTelAPI.attributeString('k', '')` threw `ArgumentError`
- `OTelAPI.attributeStringList('k', [])` threw `ArgumentError`
- `Attributes.fromJson({'key': []})` silently dropped the empty list
- `Attributes.of({'key': []})` silently dropped the empty list

### Source
- **`attribute.dart`**: Removed the two throw blocks in `Attribute._()` that rejected empty strings and empty lists.
- **`otel_api_factory.dart`**: Removed `value.isNotEmpty` guards on string and list values in `attrsFromMap`. Empty values now flow through to the constructor normally. Updated doc comment.
- **`attributes.dart`**: Removed the `value.isNotEmpty` guard and the empty-list `report()` branch in `fromJson`. Untyped empty lists now default to `List<String>` (since `.every()` returns `true` for all predicates). Updated doc comment.

### Tests
- Flipped 4 existing tests that encoded the buggy behavior (2 in `attribute_validation_test.dart`, 1 in `attribute_test.dart`, 1 in `attributes_test.dart`).
- Added 6 new regression tests: empty string stored, empty typed list stored, empty untyped list stored as `List<String>`, empty list equality, empty string equality, and `fromJson` round-trip for empty values.

### Documentation
- Added CHANGELOG entry under `[1.0.0-rc.3-wip]`.

## Verification

- `dart analyze --fatal-infos` — 0 issues
- `dart test` — 1176/1176 pass
- Reproduction script confirms all 5 empty-value cases now store correctly instead of throwing

## Scope

No behavioral change for non-empty values. The only change is that empty strings and empty lists are now accepted and stored, which aligns with the OTel spec requirement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,14 @@ fixes. See the `1.0.0-rc.3` entry for detail.
   schema URL. Both are `null` when you omit them.
   ([#108](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/108))
 
+### Fixed (spec compliance)
+- **Empty string and empty array attribute values are now stored** (api#80).
+  The OTel specification requires that empty strings and empty arrays be
+  valid attribute values. Previously `attributeString('k', '')` and
+  `attributeStringList('k', [])` threw `ArgumentError`, and
+  `Attributes.fromJson` silently dropped empty lists. Now all three paths
+  store empty values correctly.
+
 ## [1.0.0-rc.2] - 2026-08-23
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `OTelAPI.spanEvent('')`, `addEvent`, `addEventNow` and `addEvents` no longer
   throw an `ArgumentError`, per error-handling.md
   ([#117](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/117)).
+- Empty string and empty array attribute values no longer throw `ArgumentError`
+  ([#103](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/103)).
 
 ## [1.0.0-rc.3] - 2026-08-27
 
@@ -56,14 +58,6 @@ fixes. See the `1.0.0-rc.3` entry for detail.
 - `getTracer`, `getLogger` and `getMeter` no longer invent a scope version or
   schema URL. Both are `null` when you omit them.
   ([#108](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/108))
-
-### Fixed (spec compliance)
-- **Empty string and empty array attribute values are now stored** (api#80).
-  The OTel specification requires that empty strings and empty arrays be
-  valid attribute values. Previously `attributeString('k', '')` and
-  `attributeStringList('k', [])` threw `ArgumentError`, and
-  `Attributes.fromJson` silently dropped empty lists. Now all three paths
-  store empty values correctly.
 
 ## [1.0.0-rc.2] - 2026-08-23
 
--- a/lib/src/api/common/attributes.dart
+++ b/lib/src/api/common/attributes.dart
@@ -54,31 +54,28 @@ class Attributes {
       } else if (value is List<double>) {
         attributes.add(AttributeCreate.create<List<double>>(key, value));
       } else if (value is List) {
-        // Try to convert the list to a supported type
-        if (value.isNotEmpty) {
-          if (value.every((e) => e is String)) {
-            attributes.add(AttributeCreate.create<List<String>>(
-                key, value.cast<String>()));
-          } else if (value.every((e) => e is bool)) {
-            attributes.add(
-                AttributeCreate.create<List<bool>>(key, value.cast<bool>()));
-          } else if (value.every((e) => e is int)) {
-            attributes
-                .add(AttributeCreate.create<List<int>>(key, value.cast<int>()));
-          } else if (value.every((e) => e is double || e is int)) {
-            // Convert all to double
-            attributes.add(AttributeCreate.create<List<double>>(
-                key,
-                value
-                    .map((e) => e is int ? e.toDouble() : e as double)
-                    .toList()));
-          } else {
-            OTelErrorHandling.report(ArgumentError(
-                'Ignoring attribute $key because the list contains unsupported types. Only String, bool, int, double lists are allowed by the OTel specification.'));
-          }
+        // Try to convert the list to a supported type.
+        // For empty lists, .every() returns true for all predicates,
+        // so String wins by position — a reasonable default.
+        if (value.every((e) => e is String)) {
+          attributes.add(
+              AttributeCreate.create<List<String>>(key, value.cast<String>()));
+        } else if (value.every((e) => e is bool)) {
+          attributes
+              .add(AttributeCreate.create<List<bool>>(key, value.cast<bool>()));
+        } else if (value.every((e) => e is int)) {
+          attributes
+              .add(AttributeCreate.create<List<int>>(key, value.cast<int>()));
+        } else if (value.every((e) => e is double || e is int)) {
+          // Convert all to double
+          attributes.add(AttributeCreate.create<List<double>>(
+              key,
+              value
+                  .map((e) => e is int ? e.toDouble() : e as double)
+                  .toList()));
         } else {
           OTelErrorHandling.report(ArgumentError(
-              'Ignoring attribute $key because empty lists are not allowed by the OTel specification.'));
+              'Ignoring attribute $key because the list contains unsupported types. Only String, bool, int, double lists are allowed by the OTel specification.'));
         }
       } else {
         OTelErrorHandling.report(ArgumentError(
@@ -324,7 +321,7 @@ class Attributes {
 /// Extension to create Attributes from a simple Map
 extension AttributesExtension on Map<String, Object> {
   /// Convert this map to Attributes
-  /// Empty string are not allowed and are skipped
+  /// Empty strings and empty lists are stored per the OTel spec
   Attributes toAttributes() {
     return OTelFactory.getOrCreateDefault().attributesFromMap(this);
   }
--- a/lib/src/api/common/attributes.dart
+++ b/lib/src/api/common/attributes.dart
@@ -54,9 +54,9 @@ class Attributes {
       } else if (value is List<double>) {
         attributes.add(AttributeCreate.create<List<double>>(key, value));
       } else if (value is List) {
-        // Try to convert the list to a supported type.
-        // For empty lists, .every() returns true for all predicates,
-        // so String wins by position — a reasonable default.
+        // Untyped lists (List<Object> / List<dynamic>): element-check
+        // rather than hard-cast.  Empty untyped lists take the
+        // List<String> fallback because .every() is vacuously true.
         if (value.every((e) => e is String)) {
           attributes.add(
               AttributeCreate.create<List<String>>(key, value.cast<String>()));
--- a/lib/src/api/common/attributes.dart
+++ b/lib/src/api/common/attributes.dart
@@ -75,7 +75,7 @@ class Attributes {
                   .toList()));
         } else {
           OTelErrorHandling.report(ArgumentError(
-              'Ignoring attribute $key because the list contains unsupported types. Only String, bool, int, double lists are allowed by the OTel specification.'));
+              'Ignoring attribute $key because the list contains unsupported types. Only lists of Strings, bools, ints and doubles are allowed by the OTel specification.'));
         }
       } else {
         OTelErrorHandling.report(ArgumentError(
--- a/lib/src/api/factory/otel_api_factory.dart
+++ b/lib/src/api/factory/otel_api_factory.dart
@@ -161,7 +161,7 @@ class OTelAPIFactory extends OTelFactory {
   /// Creates Attributes from a map of string keys to arbitrary values.
   ///
   /// This method handles converting various value types to appropriate attribute values:
-  /// - String values become string attributes (empty strings are ignored)
+  /// - String values become string attributes (empty strings are stored per the OTel spec)
   /// - int, double, and bool values become their respective attribute types
   /// - DateTime values are converted to ISO8601 string attributes
   /// - Attribute values are passed through directly
@@ -171,9 +171,7 @@ class OTelAPIFactory extends OTelFactory {
     final attributes = <Attribute>[];
     namedMap.forEach((key, value) {
       if (value is String) {
-        if (value.isNotEmpty) {
-          attributes.add(AttributeCreate.create<String>(key, value));
-        }
+        attributes.add(AttributeCreate.create<String>(key, value));
       } else if (value is int) {
         attributes.add(AttributeCreate.create<int>(key, value));
       } else if (value is double) {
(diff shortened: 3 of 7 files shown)


lib/src/api/factory/otel_api_factory.dart
```diff
@@ -189,27 +187,25 @@ class OTelAPIFactory extends OTelFactory {
...
-            attributes.add(AttributeCreate.create<List<double>>(
-                key,
-                value
-                    .map((e) => e is int ? e.toDouble() : e as double)
-                    .toList()));
-          } else {
-            OTelErrorHandling.report(ArgumentError(
-                'Ignoring attribute $key because the list contains unsupported types. Only String, bool, int, double lists are allowed by the OTel specification.'));
-          }
+        if (value.every((e) => e is String)) {
```
robert-northmind, line 198: I think there might be an issue here now that the isNotEmpty guard is gone.

Empty lists now fall into this element-check cascade. And `every()` returns true for an empty list, since there is no element that could make it false. So the first branch always wins, and any typed empty list gets stored as a List<String>.

I ran it to double check:

```dart
Attributes.of({'k': <int>[]});  // stored as CastList<int, String>
Attributes.of({'k': <int>[]}).getIntList('k');  // Bad state: Value for key "k" is not of type List<int>
```

Same for <bool>[] and <double>[]. On main these returned null, since the entry was just dropped. So I guess this turns a silent drop into a throw, which might be the opposite of what this PR wants to fix :)

Attributes.fromJson does not have this problem, since it checks value is List<int> and other types before falling back to element probing (attributes.dart L48-55). This one goes straight to probing.

And I think the SDK inherits this one too. OTelSDKFactory extends OTelAPIFactory and does not override attributesFromMap, so the wrong element type would end up in the OTLP exporter.

Should we just mirror what fromJson does?
Something like:
```dart
} else if (value is List<String>) {
  attributes.add(AttributeCreate.create<List<String>>(key, value));
} else if (value is List<bool>) {
  attributes.add(AttributeCreate.create<List<bool>>(key, value));
} else if (value is List<int>) {
  attributes.add(AttributeCreate.create<List<int>>(key, value));
} else if (value is List<double>) {
  attributes.add(AttributeCreate.create<List<double>>(key, value));
} else if (value is List) {
  // existing element-check fallback, for List<Object>/List<dynamic>
```

Non-empty lists would behave exactly the same with this, so it is only the empty ones that change. What do you think?

lib/src/api/common/attributes.dart
```diff
@@ -54,31 +54,28 @@ class Attributes {
...
-                value
-                    .map((e) => e is int ? e.toDouble() : e as double)
-                    .toList()));
-          } else {
-            OTelErrorHandling.report(ArgumentError(
-                'Ignoring attribute $key because the list contains unsupported types. Only String, bool, int, double lists are allowed by the OTel specification.'));
-          }
+        // Try to convert the list to a supported type.
+        // For empty lists, .every() returns true for all predicates,
+        // so String wins by position — a reasonable default.
```
robert-northmind: Just a small comment. I think it is a bit more pessimistic than what this file actually does? :)

The typed checks above (L48-55) already catch <int>[], <bool>[] etc. So String only wins here for List<Object>/List<dynamic>. The "wins by position" part is only true for the untyped case I think.

Maybe something like:
```
// Only untyped lists reach here; the checks above keep a typed
// list's element type. every() returns true for an empty list, because
// there is no element that could make it false, so an empty untyped
// list takes the List<String> fallback.
```

test/unit/api/common/attribute_validation_test.dart
```diff
@@ -76,5 +79,34 @@ void main() {
...
+
+    test('Attributes.of preserves empty string', () {
+      final attrs = Attributes.of({'key': ''});
+      expect(attrs.getString('key'), equals(''));
+    });
+
+    test('Attributes.of preserves typed empty list', () {
+      final attrs = Attributes.of({'key': <String>[]});
+      expect(attrs.getStringList('key'), equals(<String>[]));
+    });
```
robert-northmind: I think this one only covers `<String>[]`, which is the one element type where the empty-list problem is invisible. 
If it used `<int>[]` I think it would have caught the `attrsFromMap` thing I commented on.

Should we cover all four?

```dart
test('Attributes.of preserves the element type of typed empty lists', () {
  expect(Attributes.of({'k': <String>[]}).getStringList('k'), equals(<String>[]));
  expect(Attributes.of({'k': <bool>[]}).getBoolList('k'), equals(<bool>[]));
  expect(Attributes.of({'k': <int>[]}).getIntList('k'), equals(<int>[]));
  expect(Attributes.of({'k': <double>[]}).getDoubleList('k'), equals(<double>[]));
});
```

test/unit/api/common/attributes_test.dart
```diff
@@ -481,12 +481,28 @@ void main() {
...
+        'emptyList': <String>[],
+      };
+      final attrs = Attributes.fromJson(json);
+
+      expect(attrs.getString('emptyStr'), equals(''));
+      expect(attrs.getStringList('emptyList'), equals(<String>[]));
+
+      final roundTripped = Attributes.fromJson(attrs.toJson());
+      expect(roundTripped.getString('emptyStr'), equals(''));
+      expect(roundTripped.getStringList('emptyList'), equals(<String>[]));
```
robert-northmind, line 509: I did not fully follow this one, but could also be me ☺️

I think `toJson()` here just copies the same Dart objects into a map, so the list is still a real `List<String>` when it goes back into fromJson. Nothing is actually serialized, so I am not sure this test would catch a JSON problem?

If we encode and decode for real:
```dart
final decoded = jsonDecode(jsonEncode(attrs.toJson())) as Map<String, dynamic>;
final roundTripped = Attributes.fromJson(decoded);
```

then `[]` comes back as `List<dynamic>`, I think, and takes the `List<String>` fallback. Still fine for this test since it uses `<String>[]`. But it means an empty `<int>[]` can never survive a real round trip, since JSON has nothing to store the element type in.

Should we maybe add a small test that documents that? So it is a known limitation instead of a surprise later.

robert-northmind (commented):
Hi @abidiahmedcom, 
I think the core change looks great to me, removing the throws from `Attribute._()` is exactly what #80 asked for. I also ran it locally, analyzer is clean and the 1176 tests pass.

I think there is one thing we should sort out before merging though. Removing the `isNotEmpty` guard in `attrsFromMap` means empty lists now hit the element-probing cascade, and every() returns true for an empty list. So typed empty lists end up as `List<String>` and the typed getters could then throw. 
On main those calls returned null, so it sort of swaps a silent drop for a throw. 

The rest are just smaller comments.

What do you think?

lib/src/api/factory/otel_api_factory.dart
abidiahmedcom: Good catch, you are right. I mirrored the `fromJson` pattern with explicit `is List<String>`, `is List<bool>`, `is List<int>`, `is List<double>` checks before the element-probing cascade. The typed empty lists now preserve their element type, and only untyped `List<Object>`/`List<dynamic>` reach the `every()` fallback. Updated the comment there too.

lib/src/api/common/attributes.dart
abidiahmedcom: Updated, thanks. The comment now says "Untyped lists" and notes that empty untyped lists take the `List<String>` fallback because `.every()` is vacuously true.

test/unit/api/common/attribute_validation_test.dart
abidiahmedcom: Done. Replaced the single `<String>[]` test with a combined test that covers all four types: `<String>[]`, `<bool>[]`, `<int>[]`, and `<double>[]`, each verified with its matching getter.

test/unit/api/common/attributes_test.dart
abidiahmedcom: Good point. Changed the round-trip test to use real `jsonEncode`/`jsonDecode` instead of just `toJson()`. Also added a separate test that documents the known limitation: empty `<int>[]` comes back as `List<String>` after a real JSON round-trip because JSON has no way to store element types.

lib/src/api/factory/otel_api_factory.dart
robert-northmind: Nice, this looks great now 👍
But I think this now surfaces something else that I missed before. 
Now that we actually store the typed empty list, a wrong-type read on it throws instead of returning null.

For example, if you do this:
```dart
Attributes.of({'k': <int>[]}).getStringList('k');
// this PR now: throws StateError: Value for key "k" is not of type List<String>
```

Before, the isNotEmpty guard dropped the list, so the key was just missing and the getter returned null. Now the key is there holding a List<int>, so `_getTyped` throws.

So I think we sort of add a new throwing path here.
And the getter docs for the `getStringList` already promise the other thing:
`/// Returns null if the key doesn't exist or if the value is not a String List.`

Here is the potential issue in the `_getTyped` in `attributes.dart`
```dart
T? _getTyped<T>(String key) {
    final attribute = _entries[key];
    if (attribute == null) return null;

    // Ensure the value matches the expected type `T`
    if (attribute.value is T) {
      return attribute.value as T;
    } else {
      // Here we throw now
      throw StateError('Value for key "$key" is not of type $T');
    }
  }
```

I think we need to update this to something like:

```dart
T? _getTyped<T>(String key) {
    final attribute = _entries[key];
    if (attribute == null) return null;

    // Ensure the value matches the expected type `T`
    if (attribute.value is T) {
      return attribute.value as T;
    }
    OTelErrorHandling.report(
        StateError('Value for key "$key" is not of type $T'));
    return null;
  }
```

The `OTelErrorHandling` is already imported here and we use it a bit further up, so no new import. And the getter docs already say null, so they would just become true.

But with this change, we also need to update some tests I think in the `attributes_test.dart` and the `otel_api_factory_test.dart`.

lib/src/api/common/attributes.dart
michaelbushe: @robert-northmind that last find is real, and it is already tracked: #81, "Common: typed Attributes getters throw StateError on a type mismatch." So this thread can resolve rather than grow, and #103 does not need to absorb it.

Worth being precise about the relationship, because it affects how we sequence the two. #103 does not create the throwing path. `_getTyped` throws on any type mismatch today, and this PR does not touch it (no changes to that method in the diff). What changes is reachability: while empty lists were silently dropped, the key was absent and the getter returned null, so an empty typed list could never reach the mismatch branch. Storing them, which #80 requires, makes an existing bug reachable on one more path.

The contract argument in #81 is the same one you made here. `getStringList` documents "Returns null if the key doesn't exist or if the value is not a String List," and the implementation throws instead, so the method contradicts its own doc comment. Underneath that, `error-handling.md` says API methods must not throw unhandled exceptions when used incorrectly by end users, which is what a wrong-type read is.

The fix is for `_getTyped` to return null on a mismatch and report through `OTelErrorHandling`, so the value is still surfaced to a strict-mode handler rather than swallowed. That covers all eight getters, not just the list ones. I am opening a follow-up PR for #81 based on this branch so the two can be reviewed in order, and I will link it here.

Two other things while I am here.

The element-type preservation you pushed for is better than what we had drafted internally. Our own reference patch for #80 type-erased every empty list to `List<String>`, on the reasoning that OTLP's empty `ArrayValue` carries no element type anyway, so the wire format is identical either way. That reasoning is correct about the wire and wrong about the in-process API, where `Attributes.of({'k': <int>[]}).getIntList('k')` returning `[]` is plainly better than returning null. Good catch, and we will fold it back into our notes.

Second, the point about the original test only covering `<String>[]` is the most valuable comment in this thread. That is the one element type where the bug is invisible, so the test would have gone green while the defect shipped. A test that cannot fail is worse than no test, because it converts an open question into a false assurance. Covering all four types was exactly right.

One process note for whoever merges: CI has not run on this PR at all, the checks list is empty, so it needs a run before merge.

lib/src/api/factory/otel_api_factory.dart
```diff
@@ -185,31 +183,37 @@ class OTelAPIFactory extends OTelFactory {
...
-                    .map((e) => e is int ? e.toDouble() : e as double)
-                    .toList()));
-          } else {
-            OTelErrorHandling.report(ArgumentError(
-                'Ignoring attribute $key because the list contains unsupported types. Only String, bool, int, double lists are allowed by the OTel specification.'));
-          }
+        // Untyped lists (List<Object> / List<dynamic>): element-check
+        // rather than hard-cast.  Empty untyped lists take the
+        // List<String> fallback because .every() is vacuously true.
+        if (value.every((e) => e is String)) {
```
michaelbushe, line 198: @robert-northmind, on the point you raised in the thread just above, that storing a typed empty list turns a wrong-type read from a silent null into a `StateError`: that is real, and it is already tracked as #81, "Common: typed Attributes getters throw StateError on a type mismatch." So this thread can resolve rather than grow, and #103 does not need to absorb it.

**#103 does not create the throwing path.** `_getTyped` throws on any type mismatch today, and this PR does not touch that method at all. What changes is reachability. While empty lists were silently dropped the key was absent, so the getter returned null and an empty typed list could never reach the mismatch branch. Storing them, which #80 requires, makes an existing bug reachable on one more path. It is a pre-existing defect becoming visible, not a regression this PR introduces.

The contract argument in #81 is the same one you made. `getStringList` documents "Returns null if the key doesn't exist or if the value is not a String List," and the implementation throws instead, so the method contradicts its own doc comment. Underneath that, `error-handling.md` says API methods must not throw unhandled exceptions when used incorrectly by end users, and a wrong-type read is exactly that.

The fix is for `_getTyped` to return null on a mismatch and report through `OTelErrorHandling`, so a strict-mode handler still surfaces it rather than having it swallowed. That covers all eight getters, not only the list ones. I have opened #106 for it, based on this branch so the two review in order. It should merge after this one.

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.3-wip]
 
+### Fixed (spec compliance)
+- **Empty string and empty array attribute values are now stored** (api#80).
```
michaelbushe: Leave Issue numbers out of the CHANGELOG, it's internal, not external.  
Shorten this whole thing to:  "Empty string and empty array attribute no longer throw ArgumentError."
The CHANGLOG is the most read doc after the README, let's keep it terse and readable.  AI has a helluva time being brief. 

I can't discuss writing with mentioning [this movie scene](https://www.youtube.com/watch?v=gA-sEfXOaEQ) from A River Runs Through It.

michaelbushe (changes requested):

lib/src/api/common/attributes.dart
```diff
@@ -54,31 +54,28 @@ class Attributes {
...
+          // Convert all to double
+          attributes.add(AttributeCreate.create<List<double>>(
+              key,
+              value
+                  .map((e) => e is int ? e.toDouble() : e as double)
+                  .toList()));
         } else {
           OTelErrorHandling.report(ArgumentError(
-              'Ignoring attribute $key because empty lists are not allowed by the OTel specification.'));
+              'Ignoring attribute $key because the list contains unsupported types. Only String, bool, int, double lists are allowed by the OTel specification.'));
```
michaelbushe: Clarity "Only lists of Strings, bools, ints and doubles are allowed" (sounds like a "double list").

michaelbushe (changes requested):

CHANGELOG.md
robert-northmind: Thanks for the input Michael. And funny movie clip 😆 that will stay in my head now

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.3-wip]
 
+### Fixed (spec compliance)
```
yuzurihaaa, line 10: You may want to rebase this from latest main https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/releases#release-v1.0.0-rc.3 is already release. Latest main wip is [1.0.0-rc.4](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/blob/main/CHANGELOG.md#100-rc4-wip)

michaelbushe: Done

lib/src/api/factory/otel_api_factory.dart
michaelbushe: Fixed by #106, merged just now as d7802d4. Your exact example, run against the branch before merge:

```
getStringList on an empty <int>[] -> null
reported: 1 -> Bad state: Attribute value for key "k" is a List<int>, not a List<String>; returning null.
getIntList (correct type)         -> []
```

Returns null instead of throwing, and the mismatch still reaches the error handler so a strict-mode handler sees it. Mutation-checked too: restoring the throw fails four tests, so the tests hold the fix rather than just sitting next to it.

Good catch. #103 made an existing bug reachable on one more path and you spotted it in review, which is exactly where it should have been caught.

PR #105, opened 2026-08-27
fix: replace enabled getter with isEnabled() method across Trace, Logs, and Metrics APIs

### LINKED ISSUE #85: [BUG] Trace API: Tracer Enabled is a getter and cannot take parameters (API surface)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#116** — fix the API first; the SDK issue stays open and tracks the SDK-side work, which depends on this landing in an API release.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Trace API, Enabled
**Requirement level:** MUST / SHOULD
**URL:** https://opentelemetry.io/docs/specs/otel/trace/api/#enabled
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#enabled

> There are currently no required parameters for this API. Parameters can be
> added in the future, therefore, the API MUST be structured in a way for
> parameters to be added.
>
> The returned value is not always static, it can change over time. The API
> SHOULD be documented that instrumentation authors needs to call this API each
> time they [create a new `Span`](https://opentelemetry.io/docs/specs/otel/trace/api/#span-creation) to ensure they have the most
> up-to-date response.

## The problem

There are two faults.

First, `enabled` is a getter. A Dart getter cannot take parameters. The spec expects future parameters, for example a span kind or a context. To add one, a maintainer must replace the getter with a method. That change breaks callers. The API is therefore not structured for new parameters.

Second, the doc comment does not tell instrumentation authors to call the API each time they create a span. The value can change over time.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/trace/tracer.dart:60`

```dart
/// Returns true if the tracer is enabled and will create sampling spans.
/// This should be checked before performing expensive operations to create spans.
bool get enabled => false;
```

## How to correct it

Replace the getter with a method. Use optional named parameters, so a new parameter does not break callers. Complete the documentation.

```dart
/// Returns whether this tracer is enabled for the provided arguments.
/// The returned value can change over time; instrumentation authors
/// need to call this each time they create a new span to ensure they
/// have the most up-to-date response.
bool enabled() => false;   // future: bool enabled({SpanKind? kind, Context? context})
```

CAUTION: The change from a getter to a method breaks callers. Add the `breaking-change` label if triage confirms the violation.

NOTE: The Metrics API and the Logs API have the same fault. See the Metrics API and Logs sections of #95.

## Notes

- Source finding files: `fable-ultra/trace-api-08-enabled-getter-not-extensible.md`, `fable-medium/trace-api-enabled-getter-not-parameterizable.md`
- Parent issue: #95

### LINKED ISSUE #90: [BUG] Logs API: Logger Enabled accepts no Context, SeverityNumber, or EventName (API surface)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#185** — fix the API first; the SDK issue stays open and tracks the SDK-side work, which depends on this landing in an API release.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Logs API, Enabled
**Requirement level:** SHOULD
**URL:** https://opentelemetry.io/docs/specs/otel/logs/api/#enabled
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/logs/api.md#enabled

> The API SHOULD accept the following parameters:
>
> - The [Context](https://opentelemetry.io/docs/specs/otel/context/) to be associated with the `LogRecord`.
>   When implicit Context is supported, then this parameter SHOULD be optional and
>   if unspecified then MUST use current Context.
> - [Severity Number](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-severitynumber) (optional)
> - [Event Name](https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-eventname) (optional)

## The problem

`Enabled` is a parameterless getter in the API and in the SDK.

None of the three parameters can reach it. A caller cannot ask whether the logger is enabled for a given severity, a given event name, or a given context.

A Dart getter also cannot take a parameter, so the shape blocks any later addition.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/logs/logger.dart:41-42`

```dart
/// Returns true if the logger is enabled.
bool get enabled => false;
```

The SDK getter is at `dartastic_opentelemetry/lib/src/logs/logger.dart:76-81`.

## How to correct it

Replace the getter with a method. Accept an optional context, an optional severity number, and an optional event name.

Default the context to the current context when the caller gives none.

CAUTION: The change from a getter to a method breaks callers. Add the `breaking-change` label if triage confirms the violation.

NOTE: #116 and #147 are the same shape problem in the Trace API and the Metrics API.

## Notes

- Source finding files: `fable-ultra/logs-11-enabled-accepts-no-parameters.md`, `fable-medium/logs-api-enabled-missing-parameters.md`, `grok/logs/002-enabled-api-parameters.md`
- Parent issue: #95

### LINKED ISSUE #99: [BUG] Metrics API: Enabled is a getter and cannot take parameters (API surface)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#147** — fix the API first; the SDK issue stays open and tracks the SDK-side work (isEnabled overrides on the SDK instruments), which depends on this landing in an API release.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Metrics API, Enabled
**Requirement level:** MUST / SHOULD
**URL:** https://opentelemetry.io/docs/specs/otel/metrics/api/#enabled
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/api.md#enabled

> There are currently no required parameters for this API. Parameters can be
> added in the future, therefore, the API MUST be structured in a way for
> parameters to be added.

## The problem

Each synchronous instrument exposes `Enabled` as a Dart getter. A getter cannot take parameters.

To add a parameter later, a maintainer must replace the getter with a method, or add a second method next to it. Both options break the API or make it inconsistent.

The value is also a `final bool` captured when the code builds the instrument. The spec says the value can change over time. A final field cannot change.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/metrics/counter.dart:57`

```dart
/// Returns whether this counter is enabled.
bool get enabled => _enabled;
```

The same getter exists at `up_down_counter.dart:57`, `histogram.dart:58`, and `gauge.dart:57`.

## How to correct it

1. Replace the getter with a method, for example `bool isEnabled()`. Use optional named parameters, so a new parameter does not break callers.
2. Document that the result can change, so instrumentation authors call it before each measurement.
3. Let an SDK subclass compute the value. Do not store it in a final field.

CAUTION: The change from a getter to a method breaks callers. Add the `breaking-change` label if triage confirms the violation.

NOTE: #116 is the same fault in the Trace API. The Logs API has it too.

## Notes

- Source finding files: `fable-ultra/metrics-api-09-enabled-not-extensible.md`
- Parent issue: #95

harshitt13:
## Description
Replaces the `enabled` getter with an `isEnabled()` method on `APITracer`, `APILogger`, `APIInstrument`, and all seven instrument subclasses (Counter, UpDownCounter, Histogram, Gauge, ObservableCounter, ObservableUpDownCounter, ObservableGauge). A Dart getter cannot take parameters, so the previous shape blocked the spec-mandated ability to add parameters (e.g. SpanKind/Context for Trace; Context/SeverityNumber/EventName for Logs) without another breaking change later.

The `enabled` constructor parameter is also removed from all instrument constructors and `*Create.create()` factories — per #99's guidance, the value should be computed by an SDK subclass on each call rather than captured once at construction time in a `final` field.

**BREAKING CHANGE**: `.enabled` → `.isEnabled()` everywhere above. Any downstream code reading `.enabled` or passing `enabled:` to these constructors will need to update.

**Explicitly out of scope**: `APITracerProvider.enabled`, `APIMeterProvider.enabled`, `APILoggerProvider.enabled` (getter *and* setter) are untouched. These are provider lifecycle/shutdown state, not the spec's per-call "Enabled" check, and are a different concept.

## Link to tracking issue
Fixes #85
Fixes #90
Fixes #99

## Testing
- `dart analyze`: 0 issues
- `dart test`: full suite passes
- `instrument_test.dart` rewritten (not just renamed) to test the new semantics: base API `isEnabled()` returns `false` by default, and a subclass overriding `isEnabled()` is respected — the actual point of the fix, since real enablement logic now lives entirely in SDK subclasses.
- All other affected test files (`logger_test.dart`, `counter_test.dart`, `gauge_test.dart`, `histogram_test.dart`, `up_down_counter_test.dart`, `observable_counter_test.dart`, `observable_gauge_test.dart`,  `observable_up_down_counter_test.dart`, `meter_test.dart`) updated to call `.isEnabled()` instead of reading `.enabled`.
- Verified via `grep -rn "\.enabled\b|get enabled\b"` across `lib/` and `test/` that no reference to the old shape remains outside the intentionally-untouched `*Provider` classes.

## Documentation
- Doc comments on each `isEnabled()` note that the returned value can change over time and should be called before each span/measurement/log record, per the spec.
- `CHANGELOG.md` updated under `[1.0.0-rc.3-wip]` with a `### Changed` / **BREAKING** entry.

## Spec Compliance
- Trace API — Enabled (MUST/SHOULD, v1.60.0): https://opentelemetry.io/docs/specs/otel/trace/api/#enabled
- Logs API — Enabled (SHOULD, v1.60.0): https://opentelemetry.io/docs/specs/otel/logs/api/#enabled
- Metrics API — Enabled (MUST/SHOULD, v1.60.0): https://opentelemetry.io/docs/specs/otel/metrics/api/#enabled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.3-wip]
 
+### Changed
+- **BREAKING**: Replaced the `enabled` getter with `isEnabled()` method on `APITracer`, `APILogger`, `APIInstrument` and all instrument subclasses. The `enabled` parameter has also been removed from all metric instrument creation API constructors and factories. This change allows SDKs to dynamically re-evaluate instrument enablement and enables adding parameters to `isEnabled()` in the future without breaking the API.
+
 ## [1.0.0-rc.2] - 2026-08-23
 
 ### Added
--- a/lib/src/api/trace/tracer.dart
+++ b/lib/src/api/trace/tracer.dart
@@ -249,8 +249,8 @@ class APITracer {
       links: links,
       spanEvents: spanEvents,
       startTime: startTime,
-      isRecording:
-          isRecording ?? true && isEnabled(kind: kind, context: contextOfSpan),
+      isRecording: (isRecording ?? true) &&
+          isEnabled(kind: kind, context: contextOfSpan),
       timeProvider: timeProvider,
     );
     return apiSpan;
--- a/lib/src/api/trace/tracer.dart
+++ b/lib/src/api/trace/tracer.dart
@@ -57,9 +57,17 @@ class APITracer {
     TimeProvider? timeProvider,
   }) : timeProvider = timeProvider ?? defaultTimeProvider;
 
-  /// Returns true if the tracer is enabled and will create sampling spans.
+  /// Returns whether this tracer is enabled for the provided arguments.
   /// This should be checked before performing expensive operations to create spans.
-  bool get enabled => false;
+  ///
+  /// The returned value can change over time; instrumentation authors need
+  /// to call this each time they create a new span to ensure they have the
+  /// most up-to-date response.
+  ///
+  /// No parameters are currently required by the spec, but this is a method
+  /// (not a getter) so parameters such as [kind] and [context] can be added
+  /// later without a breaking change.
+  bool isEnabled({SpanKind? kind, Context? context}) => false;
 
   /// Gets the currently active span from the current context
   APISpan? get currentSpan => Context.current.span;
@@ -241,7 +249,8 @@ class APITracer {
       links: links,
       spanEvents: spanEvents,
       startTime: startTime,
-      isRecording: isRecording ?? true && enabled,
+      isRecording:
+          isRecording ?? true && isEnabled(kind: kind, context: contextOfSpan),
       timeProvider: timeProvider,
     );
     return apiSpan;
--- a/lib/src/api/factory/otel_api_factory.dart
+++ b/lib/src/api/factory/otel_api_factory.dart
@@ -382,7 +382,6 @@ class OTelAPIFactory extends OTelFactory {
       name: name,
       description: description,
       unit: unit,
-      enabled: false, // API implementation is always disabled
       meter: APIMeterCreate.create(name: '@api/default'),
     );
   }
@@ -394,7 +393,6 @@ class OTelAPIFactory extends OTelFactory {
       name: name,
       description: description,
       unit: unit,
-      enabled: false, // API implementation is always disabled
       meter: APIMeterCreate.create(name: '@api/default'),
     );
   }
@@ -405,7 +403,6 @@ class OTelAPIFactory extends OTelFactory {
       name: name,
       description: description,
       unit: unit,
-      enabled: false, // API implementation is always disabled
       meter: APIMeterCreate.create(name: '@api/default'),
     );
   }
@@ -417,7 +414,6 @@ class OTelAPIFactory extends OTelFactory {
       name: name,
       description: description,
       unit: unit,
-      enabled: false, // API implementation is always disabled
       meter: APIMeterCreate.create(name: '@api/default'),
       boundaries: boundaries,
     );
@@ -430,7 +426,6 @@ class OTelAPIFactory extends OTelFactory {
       name: name,
       description: description,
       unit: unit,
-      enabled: false, // API implementation is always disabled
       meter: APIMeterCreate.create(name: '@api/default'),
       callback: callback,
     );
@@ -443,7 +438,6 @@ class OTelAPIFactory extends OTelFactory {
       name: name,
       description: description,
       unit: unit,
-      enabled: false, // API implementation is always disabled
       meter: APIMeterCreate.create(name: '@api/default'),
       callback: callback,
     );
@@ -456,7 +450,6 @@ class OTelAPIFactory extends OTelFactory {
       name: name,
       description: description,
       unit: unit,
-      enabled: false, // API implementation is always disabled
       meter: APIMeterCreate.create(name: '@api/default'),
       callback: callback,
     );
--- a/lib/src/api/logs/logger.dart
+++ b/lib/src/api/logs/logger.dart
@@ -37,9 +37,20 @@ class APILogger {
     this.attributes,
   });
 
-  /// Returns true if the logger is enabled.
+  /// Returns whether this logger is enabled for the provided arguments.
   /// Refer https://opentelemetry.io/docs/specs/otel/logs/api/#enabled
-  bool get enabled => false;
+  ///
+  /// The returned value can change over time; instrumentation authors need
+  /// to call this each time before emitting a LogRecord to ensure they have
+  /// the most up-to-date response.
+  ///
+  /// [context] The Context to associate with a would-be LogRecord. Defaults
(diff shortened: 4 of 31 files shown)


lib/src/api/trace/tracer.dart
```diff
@@ -241,7 +249,8 @@ class APITracer {
       links: links,
       spanEvents: spanEvents,
       startTime: startTime,
-      isRecording: isRecording ?? true && enabled,
+      isRecording:
+          isRecording ?? true && isEnabled(kind: kind, context: contextOfSpan),
```
michaelbushe: I think this line is a little off. `&&` binds tighter than `??`, so it's really `isRecording ?? (true && isEnabled(...))`, and the `true &&` isn't doing anything. So if someone passes `isRecording: true` we skip the enabled check completely.

The spec tells us which way it should go. `trace/sdk.md`, on `TracerConfig.enabled`:

> If a `Tracer` is disabled, it MUST behave equivalently to a No-op Tracer.

No-op spans are non-recording, so a disabled tracer has to beat an explicitly passed `isRecording`. That makes it:

```dart
isRecording: (isRecording ?? true) && isEnabled(kind: kind, context: contextOfSpan),
```

You didn't break this, it was `true && enabled` before, but let's fix it since you're already in here.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.3-wip]
 
+### Changed
+- **BREAKING**: Replaced the `enabled` getter with `isEnabled()` method on `APITracer`, `APILogger`, `APIInstrument` and all instrument subclasses. The `enabled` parameter has also been removed from all metric instrument creation API constructors and factories. This change allows SDKs to dynamically re-evaluate instrument enablement and enables adding parameters to `isEnabled()` in the future without breaking the API.
```
michaelbushe, line 11: Good call removing the getter rather than deprecating it. If we'd left a deprecated `enabled` delegating to `isEnabled()`, the SDK would still compile but `sdkCounter.enabled` and `sdkCounter.isEnabled()` would quietly disagree with each other, and nothing would tell us. This way every SDK override just fails to compile, which is what we want.

Heads up that it does break the SDK in about ten spots: `counter.dart:67`, `gauge.dart:47`, `histogram.dart:71`, `up_down_counter.dart:49` and `:72`, the three observables, `logs/logger.dart:63`, plus the `_meter.enabled` calls. All already tracked in SDK #116, #147 and #185.

So the order is this API release first, then the SDK moves its overrides to `isEnabled()` before its next one.  I'll try to get next release out today.

michaelbushe (changes requested):

PR #108, opened 2026-08-27
fix(api): getTracer/getLogger/getMeter must not invent a scope version or schema URL

benjaben:
## Summary

`getTracer`/`getLogger`/`getMeter` each defaulted a caller-omitted
`version`/`schemaUrl` to this package's own version constant
(`1.11.0.0`) and a schema URL the instrumented library never
declared, whenever *all three* optional parameters (`version`,
`schemaUrl`, `attributes`) were omitted. That same invented scope also
fed each provider's instance cache key, so e.g. `getTracer('x')` and
`getTracer('x', attributes: a)` were reported as differing in
version/schemaUrl in addition to attributes — parameters the caller
never touched.

`Tracer.startSpan` had its own fallback (`version ?? '1.0.0'`) when
building the span's `InstrumentationScope`, so even after removing the
provider-level defaulting, spans would still report a fabricated
version. Fixed that too, along with `InstrumentationScopeCreate.create`'s
own `'1.0.0'` default (only reachable through that call site now; its
other two callers already pass an explicit version or intentionally
use the no-op default for a synthetic `noop-tracer` scope).

- `version`/`schemaUrl` now pass through as the caller gave them across
  Trace, Logs, and Metrics; a null stays null end to end.
- Updated the existing tests that asserted the old (spec-violating)
  defaulting behavior, across all three signals.
- CHANGELOG entry kept to one line, links back to this PR.

## Test plan
- [x] `dart analyze` — no issues
- [x] `dart test` — full suite passes

Assisted-by: Claude Sonnet 5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - **BREAKING**: Replaced the `enabled` getter with `isEnabled()` method on `APITracer`, `APILogger`, `APIInstrument` and all instrument subclasses. The `enabled` parameter has also been removed from all metric instrument creation API constructors and factories. This change allows SDKs to dynamically re-evaluate instrument enablement and enables adding parameters to `isEnabled()` in the future without breaking the API.
 
+### Fixed (spec compliance)
+- `getTracer`/`getLogger`/`getMeter` no longer invent a scope version or
+  schema URL when the caller omits them
+  ([#108](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/108)).
+
 ## [1.0.0-rc.2] - 2026-08-23
 
 ### Added
--- a/lib/src/api/trace/tracer_provider.dart
+++ b/lib/src/api/trace/tracer_provider.dart
@@ -95,27 +95,19 @@ class APITracerProvider {
           'Invalid tracer name provided; using empty string as fallback.'));
     }
 
-    // Apply default values if none are provided
-    var effectiveVersion = version;
-    var effectiveSchemaUrl = schemaUrl;
-
-    // Only apply defaults if all optional parameters are missing
-    if (version == null && schemaUrl == null && attributes == null) {
-      effectiveVersion = OTelAPI.defaultServiceVersion;
-      effectiveSchemaUrl = OTelAPI.defaultSchemaUrl;
-    }
-
-    // Create a cache key based on the provided parameters.
-    final key = SignalInstanceKey(validatedName, effectiveVersion,
-        effectiveSchemaUrl, attributes, Signal.traces);
+    // Create a cache key based on the provided parameters. A caller who
+    // omits version/schemaUrl gets a tracer with a null scope version and
+    // schema URL — the API must not invent values the caller never stated.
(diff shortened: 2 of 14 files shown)


lib/src/api/trace/tracer_provider.dart
```diff
@@ -95,27 +95,19 @@ class APITracerProvider {
...
-      effectiveSchemaUrl = OTelAPI.defaultSchemaUrl;
-    }
-
-    // Create a cache key based on the provided parameters.
-    final key = SignalInstanceKey(validatedName, effectiveVersion,
-        effectiveSchemaUrl, attributes, Signal.traces);
+    // Create a cache key based on the provided parameters. A caller who
+    // omits version/schemaUrl gets a tracer with a null scope version and
+    // schema URL — the API must not invent values the caller never stated.
+    final key = SignalInstanceKey(
```
michaelbushe, line 101: Thanks for doing this, I hated this code. 

This covers Trace only. Metrics and Logs were missed when we made API issues for the SDK issues, despite what #64 says:

> NOTE: The same logic exists in `logs/logger_provider.dart:93` and in `metrics/meter_create.dart:26`. Those are separate issues in the Logs and Metrics sections of #95. Correct all three together.

Your choice: Either fold both into this PR with the equivalent test updates, or open API issues for both so they are scheduled rather than dropped. Folding them in is preferable since the change is identical in all three places. Thanks.

CHANGELOG.md
```diff
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.3-wip]
 
+### Fixed (spec compliance)
+- **`TracerProvider.getTracer` no longer invents a scope version or schema
+  URL** (api#64). A caller who omits `version`/`schemaUrl` now gets a tracer
```
michaelbushe: Leave out internals, like issues #'s, from the CHANGELOG, which is external only (I'll accept pushback on this). 

As I said in another PR, ["Make it half as long"](https://www.youtube.com/watch?v=gA-sEfXOaEQ), and then again.

AI loves these long CHANGELOG entries, they will cause headaches for users. Let's keep them terse.  The first sentence suffices, I think, feel free to add more if you think it's important for users.

michaelbushe (changes requested):
Good first PR.  Make a decision on 1 PR or 3, my mistake for not tracking the other two.

lib/src/api/trace/tracer_provider.dart
michaelbushe: Upon further review, let's link PRs and not issues, this is consistent with opentelemetry.io and other libs, like in java:
 Ignore empty baggage keys in `ImmutableBaggage.put` ([#8658](https://github.com/open-telemetry/opentelemetry-java/pull/8658))

benjaben:
Thanks for the review! Pushed an update addressing all three points:

- Folded the same fix into `logs/logger_provider.dart` and `metrics/meter_create.dart`, with equivalent test updates for both signals.
- Trimmed the CHANGELOG entry to one line and dropped the issue-number reference.
- Linked this PR from the CHANGELOG entry instead.

michaelbushe:
Thank you @benjaben  This is probably the ugliest wart, it's bother me for a while.

PR #112, opened 2026-08-28
fix(api): reject trace and span ids outside lowercase hex

### LINKED ISSUE #91: [BUG] Trace API: trace/span ID parsing accepts uppercase and signed hex (API half)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#193** — fix the API first; the SDK issue stays open and tracks the SDK-side work, which depends on this landing in an API release.

## Spec requirement

**Requirement level:** MUST

W3C Trace Context defines `trace-id` and `parent-id` as lowercase hex only. A header that holds a character outside that set is invalid, and a vendor must ignore it.

## The problem

`_parseTraceparent` checks field lengths only. It then hands the hex to the ID helpers, which call `int.tryParse` with radix 16 on each two-character slice.

`int.tryParse` accepts uppercase hex. It also accepts a sign character. `'AB'` gives 171. `'+f'` gives 15. `'-1'` gives -1, which wraps to 255 when the code stores it in a `Uint8List`. Fable verified this behavior directly.

A header such as `00-4BF92F3577B34DA6A3CE929D0E0E4736-00F067AA0BA902B7-01` is therefore accepted. W3C calls it invalid. The extract rules then require the Context to stay unchanged.

## Offending code

`dartastic_opentelemetry/lib/src/trace/w3c_trace_context_propagator.dart:203`

```dart
final traceId = OTel.traceIdFrom(traceIdHex);
final spanId = OTel.spanIdFrom(spanIdHex);
```

`dartastic_opentelemetry_api/lib/src/api/id/id_generator.dart:62`

```dart
final byte = int.tryParse(hexByte, radix: 16);
```

## How to correct it

Validate each field against the grammar in `_parseTraceparent`, before any conversion. Use `^[0-9a-f]{32}$` for the trace ID, `^[0-9a-f]{16}$` for the span ID, and `^[0-9a-f]{2}$` for the version and the trace flags. Return null on any mismatch.

Rejecting a character outside `[0-9a-f]` inside `IdGenerator.hexToBytes` also works.

NOTE: The trace-flags issue in this same section shares this parse path. Fix both together.

## Notes

- Source finding files: `fable-ultra/context-01-traceparent-hex-validation.md`, `fable-medium/propagation-traceparent-field-validation.md`
- Parent issue: #95

robert-northmind:
## Summary

Fixes #91. `IdGenerator.hexToBytes` accepted characters W3C Trace Context
forbids in a `trace-id` or `parent-id`.

## The violation

[§3.2.2.4](https://www.w3.org/TR/trace-context-2/): "Vendors MUST ignore the `traceparent` when the `parent-id` is invalid (for example, if it contains non-lowercase hex characters)", §3.2.2.3 says the same for `trace-id`, both spelled `HEXDIGLC` (`DIGIT / "a".."f"`). 

`hexToBytes` decoded each pair with `int.tryParse(radix: 16)`, a numeric parser that also takes uppercase, a sign and surrounding whitespace. 

It now looks each character up in `'0123456789abcdef'` and returns null for anything absent from it.

**BREAKING**: uppercase previously decoded to the *correct* bytes, so code
passing uppercase ids worked and now throws `FormatException`.

## Validation

- `dart analyze --fatal-infos` / `dart format` — clean
- `dart test` — 1185 passing


test/unit/api/trace/span_id_test.dart
```diff
@@ -44,6 +44,22 @@ void main() {
       );
     });
 
+    test('rejects a correctly sized uppercase hex string', () {
+      expect(
+        () => OTelAPI.spanIdFrom('00F067AA0BA902B7'),
+        throwsA(isA<FormatException>()),
```
robert-northmind, line 50: On the throwing logic here:
So this change make it so that we end up throwing now in the `OtelAPI.span/traceIdFrom`.
I was not sure if I should fix that as part of this change or not. 🤔
I guess it is its own change. Since it was already throwing. So should I open a new issue for that? Your thoughtS?

test/unit/api/trace/trace_id_test.dart
```diff
@@ -46,6 +46,22 @@ void main() {
...
+
+    test('rejects a signed hex string instead of corrupting the id', () {
+      // '-b' parsed to -11 and wrapped to 0xf5, so this returned an id that
+      // did not match the string it was built from.
+      expect(
+        () => OTelAPI.traceIdFrom('-bf92f3577b34da6a3ce929d0e0e4736'),
+        throwsA(isA<FormatException>()),
+      );
+    });
+
```
benjaben, line 64: is it worth adding a test for incorrect lengths?

benjaben (approved):

harshitt13 (approved):

test/unit/api/trace/trace_id_test.dart
robert-northmind: good point :) added tests for it now.

lib/src/api/otel_api.dart
```diff
@@ -610,7 +610,8 @@ class OTelAPI {
     try {
       final bytes = IdGenerator.hexToBytes(hexString);
       if (bytes == null || bytes.length != TraceId.traceIdLength) {
-        throw FormatException('TraceId must be {$TraceId.traceIdLength} bytes');
+        throw const FormatException(
+            'TraceId must be ${TraceId.traceIdLength} bytes');
```
robert-northmind, line 614: There was a formatting bug here. The `$` was in the wrong place.
Was `{$TraceId...`, but should be: Was `${TraceId...`

lib/src/api/id/id_generator.dart
```diff
@@ -51,17 +51,21 @@ class IdGenerator {
     return true;
   }
 
-  /// Parse hex string to bytes.
+  /// Lowercase hex digits, indexed by the value each one represents.
+  static final List<int> _hexDigitCodeUnits = '0123456789abcdef'.codeUnits;
```
michaelbushe: Using indexOf over the 16-element list is a linear scan, and this runs on every inbound traceparent. Two range checks would be O(1):

if (c >= 0x30 && c <= 0x39) return c - 0x30; // 0-9
if (c >= 0x61 && c <= 0x66) return c - 0x57; // a-f

michaelbushe (changes requested):
Just one efficiency improvement.

test/unit/api/trace/span_id_test.dart
michaelbushe: Yes, it should not throw. `error-handling.md` says API methods MUST NOT throw on bad input and SHOULD provide a safe default. Separate issue though, the throw predates this PR. Please open it.

When it gets fixed: return `TraceId.invalid` / `SpanId.invalid`, report through `OTelErrorHandling`, and test that the report actually fires. Install a handler in the test and assert the error arrives, same pattern `TraceState` uses for bad entries. The SDK propagator's try/catch around these then becomes an `isValid` check, so it is an API-half / SDK-half pair like #91 / #193.

michaelbushe: Nothing can throw (I made a lot of mistakes on this).  We need to test the 
The safe default here is probably returning TraceId.invalid / SpanId.invalid and reporting through OTelErrorHandling, so we should test for those.

lib/src/api/id/id_generator.dart
robert-northmind: good point. Im on it :)

robert-northmind: done

test/unit/api/trace/span_id_test.dart
robert-northmind: Here is the new issue in the API for stopping it from throwing in these cases:
https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/issues/132

harshitt13 (approved):

michaelbushe (approved):
Good, and the boundary test is a better idea than the doc note I asked for. /, :, backtick and g are exactly the four characters adjacent to the two ranges, so the off-by-one is pinned.

PR #113, opened 2026-08-28
feat: align Metrics API surface with specification requirements

### LINKED ISSUE #86: [BUG] Metrics API: asynchronous instrument creation accepts only one callback (API surface)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#142** — fix the API first; the SDK issue stays open and tracks the SDK-side work, which depends on this landing in an API release.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Metrics API, Asynchronous Instrument API
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/metrics/api/#asynchronous-instrument-api
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/api.md#asynchronous-instrument-api

> Users can provide `callback` functions, but it is up to their discretion. Therefore, this API MUST be structured to accept a variable number of `callback` functions, including none.
>
> The API MUST support creation of asynchronous instruments by passing zero or more `callback` functions to be permanently registered to the newly created instrument.

## The problem

The three creation methods for asynchronous instruments accept one optional callback. A user can register zero callbacks or one callback. A user cannot register two.

The spec requires a variable number.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/metrics/meter.dart:164-169`

```dart
APIObservableCounter<T> createObservableCounter<T extends num>({
  required String name,
  String? unit,
  String? description,
  ObservableCallback<T>? callback,
}) {
```

The same parameter exists in `createObservableUpDownCounter` (`meter.dart:197`) and `createObservableGauge` (`meter.dart:226`). The constructors carry it too, at `observable_counter.dart:27`, `observable_up_down_counter.dart:26`, and `observable_gauge.dart:26`.

## How to correct it

Change the parameter to a list:

```dart
List<ObservableCallback<T>> callbacks = const [],
```

Keep the single `callback` parameter as well if backward compatibility matters. Register every callback that the caller provides.

## Notes

- Found by: Fable 5 (both runs), Gemini 3.1 Pro
- Source finding files: `fable-ultra/metrics-api-02-async-single-callback.md`, `fable-medium/metrics-api-async-single-callback-at-creation.md`, `gemini-fail-mindful/metrics_api_async_callbacks.md`
- Parent issue: #95

### LINKED ISSUE #87: [BUG] Metrics API: multiple-instrument asynchronous callbacks are not supported (API surface)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#143** — fix the API first; the SDK issue stays open and tracks the SDK-side work, which depends on this landing in an API release.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Metrics API, Multiple-instrument callbacks
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/metrics/api/#multiple-instrument-callbacks
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/api.md#multiple-instrument-callbacks

> Idiomatic APIs for multiple-instrument Callbacks MUST distinguish the
> instrument associated with each observed `Measurement` value.
>
> Multiple-instrument Callbacks MUST be associated at the time of
> registration with a declared set of asynchronous instruments from the same `Meter` instance.

## The problem

The API supports a callback for one instrument only. There is no way to register one callback across several instruments.

A batch callback matters when one expensive operation produces several measurements. Without it, the user repeats that operation for each instrument.

## Offending code

- `dartastic_opentelemetry_api/lib/src/api/metrics/observable_callback.dart`
- `dartastic_opentelemetry/lib/src/metrics/meter.dart:223-243`

## How to correct it

Add a batch registration method on the meter. The method takes a callback and the set of instruments.

The observe result passed to the callback must name the instrument for each measurement.

Reject an instrument that comes from a different meter instance.

## Notes

- Found by: Grok 4.5
- Source finding files: `grok/metrics/013-batch-async-callbacks-missing.md`
- Parent issue: #95

### LINKED ISSUE #88: [BUG] Metrics API: advisory parameters are not accepted by instrument creation (API surface)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#144** — fix the API first; the SDK issue stays open and tracks the SDK-side work, which depends on this landing in an API release.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Metrics API, Synchronous Instrument API
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/metrics/api/#synchronous-instrument-api
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/api.md#synchronous-instrument-api

> * `advisory` parameters associated with the instrument `kind`.
>
>   Users can provide `advisory` parameters, but its up to their discretion. Therefore, this API needs to be structured to accept `advisory` parameters, but MUST NOT obligate the user to provide it.

## The problem

Only `createHistogram` accepts an advisory parameter. That parameter is `List<double>? boundaries`, which carries the `ExplicitBucketBoundaries` advisory.

The other six creation methods accept `name`, `unit`, and `description` only. Observable instruments also accept a callback.

There is no way to pass the general `Attributes` advisory for any instrument. There is also no structure that a new advisory parameter could use.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/metrics/meter.dart:57-61`

```dart
APICounter<T> createCounter<T extends num>({
  required String name,
  String? unit,
  String? description,
}) {
```

The histogram advisory sits at `meter.dart:109-114`.

## How to correct it

Add an optional advisory parameter to all seven `create*` methods. An `attributeKeys: List<String>?` parameter, or a small `Advisory` object, both work.

Pass the value through the factory methods and the instrument constructors, so an SDK implementation can read it. Do not validate it.

NOTE: The only stable advisory parameter is `ExplicitBucketBoundaries`, and the code supports it. The general `Attributes` advisory has the status **Development**. This issue is therefore low priority.

## Notes

- Found by: Fable 5 (x-high run), Grok 4.5, Gemini 3.1 Pro
- Source finding files: `fable-ultra/metrics-api-05-advisory-params-missing.md`, `grok/metrics/003-advisory-parameters-missing.md`, `gemini-fail-mindful/metrics_api_advisory_parameters.md`
- Parent issue: #95

### LINKED ISSUE #89: [BUG] Metrics API: no-op MeterProvider holds configuration and operational state (API surface)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#145** — fix the API first; the SDK issue stays open and tracks the SDK-side work, which depends on this landing in an API release.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Metrics No-Op, MeterProvider
**URL:** https://opentelemetry.io/docs/specs/otel/metrics/noop/#meterprovider
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/noop.md#meterprovider

> The MeterProviders created by the No-Op needs to hold as small a memory footprint as possible. Therefore, all MeterProviders created MUST NOT
> hold configuration or operational state.

## The problem

`APIMeterProvider` is the no-op implementation of this package. Its own doc comment says so.

The class holds configuration. It also holds operational state.

`_endpoint`, `_serviceName`, and `_serviceVersion` are exporter and resource configuration. They do not belong on any meter provider API surface.

`_enabled`, `_isShutdown`, and `_meterCache` are operational state. `shutdown()` and `forceFlush()` change that state and branch on it.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/metrics/meter_provider.dart:28-44`

```dart
String _endpoint;
String _serviceName;
String? _serviceVersion;
bool _enabled = false; // API, SDK defaults true
bool _isShutdown;
final Map<SignalInstanceKey, APIMeter> _meterCache = {};
```

`shutdown()` is at `meter_provider.dart:149`. `forceFlush()` is at `meter_provider.dart:170`.

## How to correct it

1. Remove the endpoint, service name, and service version fields. Move them to the SDK meter provider.
2. Remove the enabled flag, the shutdown flag, and the meter cache.
3. Return a no-op meter from `getMeter`. One shared instance is permitted and costs less.

CAUTION: This change removes public state from an API class. It can break callers. Add the `breaking-change` label if triage confirms the violation.

## Notes

- Source finding files: `fable-ultra/metrics-api-03-noop-provider-holds-state.md`
- Parent issue: #95

harshitt13:
This PR aligns the `APIMeter` and its related instrument classes with the OpenTelemetry specification, addressing several missing surface areas and correcting caching behavior in the base API.

### Changes Included
- **Multiple Async Callbacks (#86):** Updated asynchronous instruments (`APIObservableCounter`, `APIObservableUpDownCounter`, `APIObservableGauge`) to support multiple callbacks. Added `addCallback` and `removeCallback` methods. `addCallback` now returns an `APICallbackRegistration` handle with an `unregister()` method for ergonomic cleanup.
- **Batch Callback Registration (#87):** Added `registerBatchCallback` to `APIMeter`, returning an `APIBatchCallbackRegistration`. Includes strict runtime validation to reject non-observable instruments or instruments belonging to a different meter instance.
- **Instrument Advisory Parameters (#88):** Added `InstrumentAdvisory` support to all 7 `create*` methods. For `createHistogram`, backward compatibility is preserved by ensuring the explicit `boundaries` parameter takes precedence over `advisory.explicitBucketBoundaries`.
- **Stateless API Providers (#89):** Refactored `APIMeterProvider` to be a pure, stateless factory rather than caching by name. Instance caching and deduplication is now strictly the responsibility of the SDK subclass.

### Quality & Testing
- Added extensive regression tests ensuring safe behavior around double-removal of callbacks, batch-registration edge cases, and advisory parameter precedence logic.

Closes #86
Closes #87
Closes #88
Closes #89


lib/src/api/metrics/meter_provider.dart
```diff
@@ -25,47 +20,11 @@ part 'meter_provider_create.dart';
...
-        _isShutdown = isShutdown;
-
-  /// Gets the endpoint URL where metrics data is sent.
-  String get endpoint => _endpoint;
-
-  /// Sets the endpoint URL where metrics data is sent.
-  set endpoint(String value) {
-    _endpoint = value;
-  }
+  APIMeterProvider._();
```
michaelbushe, line 27: This is where the SDK breaks. The SDK's `MeterProvider` stores all of its state in an `APIMeterProvider` delegate: `endpoint`, `serviceName`, `serviceVersion`, `enabled` and `isShutdown` are all `=> delegate.x`, nine references in `meter_provider.dart`. Removing them here is a compile error there, which is the good outcome. The bad one is that `isShutdown` gates `getMeter`, `collectAllMetrics`, `forceFlush` and `shutdown` in the SDK, so if someone "fixes" the compile error by dropping the delegation, the SDK keeps accepting metrics after shutdown with no error anywhere.

Same problem one layer down: the SDK uses `implements`, not `extends`, on `APIMeter`, `NoopMeter`, and all six observable instrument classes. Dart requires every interface member be declared, so `advisory`, `addCallback`, `removeCallback`, `callbacks` and `registerBatchCallback` break eight SDK classes at compile time.

For what it's worth, I'd rather have this than `@Deprecated` stubs whose setters discard writes. That version compiles and ships the broken shutdown guard with only a warning. Yours forces the fix. But it does mean this needs the SDK PR alongside it.

lib/src/api/metrics/meter.dart
```diff
@@ -112,18 +119,26 @@ class APIMeter {
...
+    InstrumentAdvisory? advisory,
   }) {
     if (name.isEmpty) {
       throw ArgumentError('Histogram name must not be empty');
     }
 
+    // Precedence: explicit boundaries parameter wins over advisory.explicitBucketBoundaries
+    final effectiveAdvisory = boundaries != null
+        ? InstrumentAdvisory(explicitBucketBoundaries: boundaries)
+        : advisory;
```
michaelbushe, line 134: Two problems here.

When `boundaries` is set, this builds a fresh `InstrumentAdvisory` and throws the caller's `advisory` away entirely. So `createHistogram(boundaries: [1, 2], advisory: InstrumentAdvisory(attributeKeys: ['route']))` silently loses `attributeKeys`. It should merge: keep the caller's advisory and override just the boundaries field.

And the CHANGELOG says the opposite of this code. Line 19 there: "When both are specified, `advisory.explicitBucketBoundaries` takes precedence." The code says `boundaries` wins, and so does the PR description. One of them needs to change. Since `boundaries` is now deprecated, having the deprecated parameter win is a bit odd, but it is the compat-preserving choice. Pick one and make the two agree.

lib/src/api/metrics/batch_callback.dart
```diff
@@ -0,0 +1,24 @@
...
+
+/// Passed to a [BatchObservableCallback] to record observations.
+///
+/// The observe method accepts [num] (not a generic T) because a single
+/// batch callback may observe across instruments with different numeric
+/// types (e.g., `Counter<int>` and `Gauge<double>`). The no-op API layer
+/// does not validate the value type; the SDK may runtime-check it.
+abstract class BatchObservableResult {
+  /// Records a measurement for [instrument] with [value] and optional [attributes].
+  void observe(dynamic instrument, num value, [Attributes? attributes]);
```
michaelbushe, line 17: `dynamic` here gives up static typing on the one thing the spec insists on. `metrics/api.md:453`: multiple-instrument callbacks "MUST distinguish the instrument associated with each observed Measurement value." This does distinguish it, but only at runtime.

The cause is that `APIObservableCounter`, `APIObservableGauge` and `APIObservableUpDownCounter` share no supertype. A marker interface, say `APIObservableInstrument`, that all three implement would let this take a real type and would also let `registerBatchCallback` reject a synchronous instrument at compile time instead of in the runtime check at line 284.

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **`InstrumentAdvisory`** parameter on all seven `APIMeter.create*` methods to allow passing advisory hints to the SDK, such as `explicitBucketBoundaries` or `attributeKeys`. ([#88](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/88))
+- **Multiple callbacks per asynchronous instrument**: `createObservableCounter`, `createObservableUpDownCounter`, and `createObservableGauge` now accept a `callbacks` list. Added `addCallback(callback)` and `removeCallback(callback)` to the resulting API observable instruments to dynamically manage callbacks. `addCallback` returns a registration handle, while `removeCallback` allows direct reference removal (providing both patterns since single-instrument callbacks are simple enough not to strictly require a handle). ([#86](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/86))
```
michaelbushe, line 14: These link `#86`, `#87`, `#88` under `/pull/`, but those are the issues. The PR is #113. Convention here is the PR link, `([#113](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/113))`, on every entry. Same for the three under Changed.

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+### Changed
+
+- **BREAKING**: `APIMeterProvider` is now completely stateless in the API layer. Calling `getMeter()` returns a fresh `APIMeter` each time, correctly preserving the meter `name`, `version`, `schemaUrl`, and `attributes` for the no-op fallback, resolving an issue where the shared singleton no-op meter failed to retain distinct configuration. ([#89](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/89))
```
michaelbushe, line 18: This is the one genuinely breaking change in the PR and the entry does not say what breaks. It describes the fresh-meter behavior but never mentions that `endpoint`, `serviceName`, `serviceVersion`, `enabled` and `isShutdown` are gone from `APIMeterProvider`. Someone reading this before upgrading has no idea `meterProvider.endpoint` will stop compiling. List the removed members.

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **BREAKING**: The `boundaries` parameter on `createHistogram` is deprecated. Use `advisory: InstrumentAdvisory(explicitBucketBoundaries: ...)` instead. When both are specified, `advisory.explicitBucketBoundaries` takes precedence. ([#88](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/88))
+- **BREAKING**: The `callback` parameter on the three `createObservable*` methods is deprecated. Use the new `callbacks` list instead. Passing a `callback` will prepend it to `callbacks`. ([#86](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/86))
```
michaelbushe, line 20: Neither of these is breaking. `boundaries` and `callback` are deprecated, not removed, and existing calls keep compiling and behaving the same. Marking a deprecation **BREAKING** makes the real one above harder to spot. Drop the label on both, or move them under a `### Deprecated` heading, which is what Keep a Changelog uses for exactly this.

michaelbushe (dismissed):
The stateless provider is the right call, and I'd take your hard removal over deprecation stubs. See the inline note on why.

But this cannot merge on its own: it does not compile against the SDK, and one of the failure modes is silent. Details on `APIMeterProvider`. There is now an `integrate/api-rc4` branch on the SDK that resolves the API from git `main`, which is where the SDK half of this belongs. Adapt the SDK there and the two can land together.

Sequencing: #125 carves out the no-op silence half of this into its own PR, and both edit `meter_provider.dart`. I'd rather take #125 first since it is one MUST in three files, then rebase this behind it. That also shrinks this PR to the surface changes, which are the part that needs the SDK work.

Two things on the CHANGELOG, inline. The links point at issues rather than this PR, and the precedence sentence for `createHistogram` says the opposite of what the code does.

Also inline: `boundaries` currently discards the caller's whole `advisory`, so `attributeKeys` is silently lost. That one is a bug, not a preference.

The `throw ArgumentError` on empty names in every `create*` is #70, not yours. Leave it.

michaelbushe:
Correcting something in my review: "this cannot merge on its own" was wrong, and if it is why this has sat still, that is on me.

The SDK is pinned to `>=1.0.0-rc.3 <1.0.0-rc.4` on its `main`, so nothing that resolves from pub.dev sees API `main` until rc.4 ships. The only consumer of API `main` is the SDK's `integrate/api-rc4` branch, which resolves it from git on purpose so the adaptation can be built there. Merging this to API `main` breaks nothing. That is exactly what the pin is for.

So the order is: land this here, adapt the SDK on `integrate/api-rc4` afterward, release rc.4 once both are done. No paired SDK PR is needed before this merges.

The inline findings still stand, and they are what is holding it: `boundaries` discarding the caller's `advisory` and losing `attributeKeys`, the CHANGELOG saying the opposite precedence from the code, the issue links, and `dynamic instrument`. Fix those and it can go in. #125 covers the no-op half, so if that lands first, rebase behind it and this gets smaller.

lib/src/api/metrics/meter_provider.dart
michaelbushe: Retracting this one. The SDK is pinned below rc.4, so merging here breaks nothing; the adaptation happens on the SDK's integrate/api-rc4 branch afterward. My framing was wrong, not your change. Your hard removal of the state is still the right call.

lib/src/api/metrics/meter.dart
michaelbushe: Fixed in ac4dd0a. boundaries still wins, and the rest of the advisory is kept, so attributeKeys survives. Test added; reverting the merge fails it.

lib/src/api/metrics/batch_callback.dart
michaelbushe: Done in ac4dd0a: APIObservableInstrument, implemented by the three observables, and observe and registerBatchCallback take it. A synchronous instrument is now a compile-time error. The foreign-meter case reports instead of throwing, per error-handling.md.

CHANGELOG.md
michaelbushe: Fixed in ac4dd0a, all entries link this PR.

CHANGELOG.md
michaelbushe: Fixed in ac4dd0a, the entry now names endpoint, serviceName, serviceVersion, enabled and isShutdown.

CHANGELOG.md
michaelbushe: Fixed in ac4dd0a, both moved under a Deprecated heading without the BREAKING label.

michaelbushe (approved):

PR #114, opened 2026-08-28
fix(api): reject trace flags outside two lowercase hex digits

### LINKED ISSUE #92: [BUG] Trace API: TraceFlags.fromString coerces invalid input to 0x00 (API half)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#194** — fix the API first; the SDK issue stays open and tracks the SDK-side work, which depends on this landing in an API release.

## Spec requirement

**Requirement level:** MUST

W3C Trace Context defines `trace-flags` as two lowercase hex digits. When any field holds a character outside the allowed set, the whole header is invalid and a vendor must not use it. The vendor restarts the trace instead.

## The problem

`_parseTraceparent` checks the length of the trace-flags field only. It then calls `TraceFlags.fromString`.

That factory swallows a parse failure and substitutes "no flags".

A header such as `00-<valid-trace-id>-<valid-span-id>-zz` is therefore accepted as a valid remote span context with flags `0x00`. W3C requires the whole header to be rejected.

Two things go wrong. The validation requirement is not met, and the sampled bit is silently dropped. A restarted trace would have decided that bit for itself.

## Offending code

`dartastic_opentelemetry/lib/src/trace/w3c_trace_context_propagator.dart:205`

```dart
final traceFlags = TraceFlags.fromString(traceFlagsHex);
```

`dartastic_opentelemetry_api/lib/src/api/trace/trace_flags.dart:48`

```dart
factory TraceFlags.fromString(String hex) {
  final flags = int.tryParse(hex, radix: 16) ?? NONE_FLAG;
  return TraceFlagsCreate.create(flags);
}
```

## How to correct it

Check `traceFlagsHex` against `^[0-9a-f]{2}$` inside `_parseTraceparent`. Return null on a mismatch. Do not rely on the lenient factory.

NOTE: Uppercase flags pass today for the same reason as the hex validation issue in this same section. One regular expression fixes both.

## Notes

- Source finding files: `fable-ultra/context-02-trace-flags-invalid-coerced.md`, `fable-medium/propagation-traceparent-field-validation.md`
- Parent issue: #95

robert-northmind:
#### Description

`TraceFlags.fromString` turns malformed input such as `zz` into `00`, so callers cannot distinguish a parse failure from valid, unsampled flags. It now returns null unless the input is exactly two lowercase hex digits, using the strict `IdGenerator.hexToBytes` decoder from merged #112.

`TraceFlagsCreate.create` also retains only the low byte of integer flags. This keeps `asByte` in range and `toString()` at two hex digits, including when `SpanContext.fromJson` constructs flags through the installed factory.

**BREAKING**: `fromString` changes from a factory constructor to a static method returning `TraceFlags?`. Callers must handle null. A propagator must reject the enclosing header when parsing fails instead of substituting no flags.

The [OpenTelemetry W3C Trace Context requirement](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#w3c-trace-context-requirements), also checked against the issue's [pinned v1.60.0 source](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/context/api-propagators.md#w3c-trace-context-requirements), says propagators:

> MUST parse and validate the `traceparent` and `tracestate` HTTP headers

[W3C §3.2.2.2](https://www.w3.org/TR/trace-context-2/#version-format) defines `trace-flags = 2 HEXDIGLC ; 8 bit flags`. [§4.1.2](https://www.w3.org/TR/trace-context-2/#a-traceparent-is-received) requires replacing the trace context and deleting incoming tracestate when the flags are invalid. Returning null exposes that failure to the propagator; the API parser itself does not decide how to restart a trace.

#### Link to tracking issue

Fixes #92 (API half).

Header rejection remains tracked in [SDK #194](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/issues/194) and [SDK PR #271](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/pull/271). SDK `main` passes the parse result to `OTel.spanContext(traceFlags: ...)`, which defaults null back to no flags. Upgrading the API alone therefore does not fix header rejection; the SDK must validate the field or explicitly reject a null result before constructing the context.

#### Testing

Rebased onto current API `main` (`fa454ff`), including merged #112. The diff contains only trace-flags changes, documentation and tests.

- The 18 focused tests cover all 256 byte values, invalid ASCII in either digit, Unicode, malformed lengths, sampling-bit preservation, integer normalization and the `SpanContext.fromJson` path. Seven of these tests fail against unmodified `main` and all pass on this branch, demonstrating the existing bugs and their fixes.
- Dart 3.12.2: `dart analyze --fatal-infos` and `dart format --output=none --set-exit-if-changed .` pass.
- Native suite: 1,240 tests pass. Chrome with dart2js and dart2wasm: 2,460 tests pass.
- Clean line coverage: 100.0000% (2,480/2,480 lines), matching the published baseline. The required pre-push hook also passes.
- All six [GitHub CI jobs](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/actions/runs/34819244955) pass: analysis, stable and beta tests, coverage, browser tests and package analysis.

#### PR Readiness Checklist

- [x] `CHANGELOG.md` updated, or not needed for this change: nullable parsing is under `### Changed` with `**BREAKING**:`, and byte normalization is under `### Fixed`; both link to #114.
- [x] Documentation is updated, or not needed for this change: `TraceFlags.fromString` documents invalid input and header rejection; `OTelAPI.traceFlags` and `TraceFlagsCreate.create` document low-byte normalization.
- [x] Changes are verified with spec compliance, or not applicable: the [OpenTelemetry propagation requirement](https://opentelemetry.io/docs/specs/otel/context/api-propagators/#w3c-trace-context-requirements) and W3C grammar and rejection rules are linked and quoted above.
- [x] `./tool/setup-hooks.sh` is installed, or `tool/coverage.sh` passes locally: hooks installed; analyzer, tests and coverage gate passed before pushing.
- [x] Assisted-by (AI tools used such as "Fable 5", or "None"): Claude Opus 5 and GPT-6, matching the commits' `Assisted-by:` trailers. The original commits also credit Cursor.
- [x] I, a human, authored this pull request and stand behind these changes.


michaelbushe:
@robert-northmind FYI - this was waiting on #112, which has been merged.

PR #115, opened 2026-08-28
fix(api): move updated/new TraceState entries to the front

### LINKED ISSUE #61: [BUG] Trace API: TraceState.put does not move new or updated entries to the front
> **Migrated from MindfulSoftwareLLC/dartastic_opentelemetry#108** — the fix belongs entirely in `dartastic_opentelemetry_api`; the SDK issue is closed as moved.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Trace API, TraceState
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/trace/api/#tracestate
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#tracestate

> These operations MUST follow the rules described in the [W3C Trace Context specification](https://www.w3.org/TR/trace-context/#mutating-the-tracestate-field).

The W3C rules for "Mutating the tracestate Field" say:

> **Update key/value pair** — The value of any key can be updated. Modified keys MUST be moved to the beginning (left) of the list.
> **Add new key/value pair** — ... added to the beginning (left) of the list.

## The problem

`put` keeps the insertion order of a Dart `Map`. A new key goes to the end of the list. An updated key stays in its old position.

`toString()` writes the entries in that order. The outgoing `tracestate` header therefore has the changed entry in the wrong position.

Position in `tracestate` carries meaning. It shows how recent an entry is, and downstream systems truncate from the right.

The overflow path has the same fault. It removes `newEntries.keys.first`. With correct left-is-newest order, that call removes the newest entry instead of the oldest one.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/trace/trace_state.dart:79`

```dart
final newEntries = Map<String, String>.from(_entries);

// If we already have this key, just update its value
if (newEntries.containsKey(key)) {
  newEntries[key] = value;          // stays at its old position
  return factory.traceState(newEntries);
}
...
newEntries[key] = value;            // appended at the END of the list
return factory.traceState(newEntries);
```

`trace_state.dart:113` serializes in that same order.

## How to correct it

Build the new map with the changed entry first:

```dart
TraceState put(String key, String value) {
  ...
  final newEntries = <String, String>{key: value};       // new/updated entry first
  _entries.forEach((k, v) {
    if (k != key && newEntries.length < _maxKeyValuePairs) newEntries[k] = v;
  });
  return factory.traceState(newEntries);
}
```

This also corrects the overflow path. Entries then drop from the end, which removes the oldest ones.

## Notes

- Source finding files: `fable-ultra/trace-api-03-tracestate-mutation-order.md`, `fable-medium/trace-api-tracestate-mutation-order.md`, `grok/trace/020-tracestate-put-order.md`
- Parent issue: #95

benjaben:
## Summary
- `TraceState.put` now moves an updated or newly-added key to the front of the entries list, matching the W3C Trace Context requirement that a mutation moves the entry to the beginning (left) of `tracestate`.
- Fixes the overflow path in the same fix: since eviction now trims from the end of the (correctly-ordered) list, it drops the oldest entry instead of the newest.
- Fixed a pre-existing coverage test that had encoded the old (incorrect) eviction direction, and added ordering/eviction assertions to the spec-compliance suite.
- Addressed review feedback: restored the eviction-rule comment, changed the eviction loop to `break` once the cap is reached, updated the stale `put()` doc comment, and added tests for updating at the 32-entry cap, updating the already-frontmost entry, and the W3C spec's own example header.

## Test plan
- [x] `dart test` — all 1179 tests pass
- [x] `dart analyze` — no issues
- [x] Test coverage includes: new key moves to front, updated key moves to front, overflow evicts the oldest (rightmost) entry rather than the newest, updating at the 32-entry cap retains all entries, updating the already-frontmost key is a no-op reorder, and the W3C spec example header (update then add)

Fixes #61


lib/src/api/trace/trace_state.dart
```diff
@@ -83,24 +83,14 @@ class TraceState {
...
-    if (newEntries.length >= _maxKeyValuePairs) {
-      // Remove the first key to make room
-      if (newEntries.isNotEmpty) {
-        final oldestKey = newEntries.keys.first;
-        newEntries.remove(oldestKey);
+    // Per W3C Trace Context, an updated or new entry moves to the front
+    // (left) of the list, so build the new map starting with it.
+    final newEntries = <String, String>{key: value};
+    for (final entry in _entries.entries) {
+      if (entry.key != key && newEntries.length < _maxKeyValuePairs) {
```
michaelbushe: You deleted the "remove the oldest entry" comment, and it was the only place the eviction rule was written down. The behavior is right, stopping here drops the tail and the tail is the oldest under the new ordering, but nothing in the code says that any more. Put a short note back on this line.

Worth using `break` instead of letting the condition go false, since once the map is full there is nothing left to do and the intent reads better.

One thing not to do here: the 32 cap is correct and stays, it is the W3C grammar limit on the list. The fuller truncation procedure, the 512-character budget and dropping entries over 128 characters first, is a separate gap I filed as #126. Do not pull that into this PR, #128 is almost done for it.

lib/src/api/trace/trace_state.dart
```diff
@@ -83,24 +83,14 @@ class TraceState {
...
-
-    // If adding a new key would exceed the limit, remove the oldest entry
-    if (newEntries.length >= _maxKeyValuePairs) {
-      // Remove the first key to make room
-      if (newEntries.isNotEmpty) {
-        final oldestKey = newEntries.keys.first;
-        newEntries.remove(oldestKey);
+    // Per W3C Trace Context, an updated or new entry moves to the front
+    // (left) of the list, so build the new map starting with it.
+    final newEntries = <String, String>{key: value};
```
michaelbushe, line 90: The doc comment on `put()` just above is stale. It still only mentions the 32 limit and "the oldest entries are removed", with nothing about the entry moving to the front, which is the whole point of this change. Anyone reading the API docs still gets the old model. Worth saying there that a new or updated entry goes to the beginning of the list per W3C, and that the rightmost entries are the oldest and go first.

Reviewing this surfaced that the 32 cap is only part of the W3C truncation rules. The rest, the 512-character budget and removing entries over 128 characters first, is now #126 here and MindfulSoftwareLLC/dartastic_opentelemetry#295 for the propagator side. Out of scope for this PR.

test/unit/api/trace/trace_state_spec_compliance_test.dart
```diff
@@ -47,6 +47,39 @@ void main() {
...
+      // (first/leftmost) is newest and vendor31 (last/rightmost) is oldest.
+      final traceState = TraceState.fromMap(entries);
+      final result = traceState.put('vendornew', 'v');
+
+      expect(result.entries.length, equals(32));
+      expect(result.get('vendornew'), equals('v'));
+      expect(result.get('vendor31'), isNull); // oldest entry evicted
+      expect(result.get('vendor0'), equals('value0')); // newest retained
+      expect(result.entries.keys.first, equals('vendornew'));
+    });
```
michaelbushe, line 80: Missing the case that actually worries me: updating an existing key when you're already at 32 entries. That must keep all 32, and it's the path where an off-by-one silently eats a real entry. I traced your loop and it's correct, it just isn't tested.

```dart
test('updating a key at the 32-entry limit does not drop any entry', () {
```

Two more worth having: updating the already-frontmost key should leave it in front, and the W3C spec's own example header, `congo=t61rcWkgMzE,rojo=00f067aa0ba902b7`, updated then added to. That last one is the closest thing to a real-world regression guard.

michaelbushe (dismissed):

lib/src/api/trace/trace_state.dart
benjaben: Updated!

test/unit/api/trace/trace_api_coverage_test.dart
```diff
@@ -135,8 +135,11 @@ void main() {
       final traceState = OTelAPI.traceState(full);
       final evicted = traceState.put('overflow', 'v');
 
+      // Map/header order is left-to-right = newest-to-oldest (W3C), so
+      // key0 (first) is newest and key31 (last) is oldest and gets evicted.
       expect(evicted.get('overflow'), equals('v'));
-      expect(evicted.get('key0'), isNull, reason: 'oldest entry is evicted');
+      expect(evicted.get('key31'), isNull, reason: 'oldest entry is evicted');
+      expect(evicted.get('key0'), equals('value0'));
```
robert-northmind, line 142: Should we maybe assert the full key order here?

```dart
expect(evicted.entries.keys.toList(), [
  'overflow',
  for (var i = 0; i < 31; i++) 'key$i',
]);
```

Then we would check that all the retained entries moved one position to the right, and that key31 is gone. Maybe this makes the expected behavior a bit clearer. 
Not a blocking comment though, just a thought :)

robert-northmind (approved):
Nice, looks good to me 👍 I just had one small suggestion on the test assertion.
I think Michael’s feedback looks addressed.
If we just resolve the CHANGELOG conflict, then I think this looks good to get merged.

test/unit/api/trace/trace_api_coverage_test.dart
michaelbushe: Applied in 21a36a6. Reads better and it is a stronger check: reverting the front-insert fails it, where the three spot checks would have too but less legibly.

michaelbushe (approved):

Issue #49, opened 2026-07-11
Propose W3C multi-tenant tracestate (tenant-id@system-id) upstream to OpenTelemetry


michaelbushe:
## The gap

The W3C Trace Context spec defines a **second key form** for `tracestate` list members: `{tenant-id}@{system-id}` — *"in the case of multi-tenant vendors, the key SHOULD be in this format"* (used in the wild, e.g. Dynatrace's `{tenant}@dt`). OpenTelemetry has **no trace of it**: the spec's TraceState API section, the semantic conventions, and the language APIs (Java, Go, JS, Python surveyed) all treat tracestate keys as opaque strings with no multi-tenant surface — some SDK validators are stricter than W3C and would reject the form outright.

## What this library already does (prior art)

- Validation accepts the form with per-part W3C grammars (tenant-id ≤ 241 starting `lcalpha/DIGIT`; system-id ≤ 14 starting `lcalpha`) — shipped since 1.0.0-beta.7.
- #48 adds first-class, vendor-neutral helpers: `TraceState.multiTenantKey()`, `putMultiTenant`/`getMultiTenant`, `tenantsForSystem()` — fail-fast validation, header round-trip tested.

## The SIG angle

This is a concrete, well-scoped upstream contribution for the **OTel Client SIG / specification**:

1. **Spec issue** (opentelemetry-specification): the TraceState API section should acknowledge the W3C multi-tenant key form and require implementations to *accept* it (some validators reject it — a W3C conformance bug), and optionally define an API for it.
2. **Proposal**: add `multiTenantKey`-style accessors to the TraceState API spec (this library's #48 surface is the strawman — vendor-neutral, no behavior beyond the W3C grammar).
3. **Positioning**: Dart's reference implementation is the first OTel API with first-class support — useful prior art for the proposal, and a good visibility point for the Dart/Flutter SIG formation.

Draft upstream issue text lives below; Michael posts upstream comms himself.

<details><summary>Draft for opentelemetry-specification</summary>

**Title:** TraceState API ignores W3C's multi-tenant key form (`tenant-id@system-id`)

[...]

PR #116, opened 2026-08-31
fix(api): TraceState.fromString rejects a repeated key

### LINKED ISSUE #93: [BUG] Trace API: TraceState parsing lacks W3C key/value validation (API half)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#195** — fix the API first; the SDK issue stays open and tracks the SDK-side work, which depends on this landing in an API release.

## Spec requirement

**Requirement level:** MUST

> A W3C Trace Context propagator MUST propagate a valid `tracestate` unless the value is empty, in which case the `tracestate` header may be omitted.

## The problem

`_parseTracestate` runs no grammar check. The API package already holds a validating parser, `TraceState.fromString`. The propagator does not use it.

Four faults follow.

1. A key that breaks the W3C grammar is accepted and stored on the extracted span context. `inject` later writes it out again, so this propagator can emit an invalid header downstream.
2. The code calls `.trim()` on the value. A leading space is legal and significant in a W3C value. Another vendor's data is therefore changed in transit.
3. A duplicate key is collapsed by the map, and the last one wins. W3C treats a duplicate key as making the header invalid.
4. The 32-member limit is not applied during parsing. The unused validating parser does apply it.

## Offending code

`dartastic_opentelemetry/lib/src/trace/w3c_trace_context_propagator.dart:241-272`

```dart
final key = trimmedEntry.substring(0, separatorIndex).trim();
final value = trimmedEntry.substring(separatorIndex + 1).trim();

if (key.isNotEmpty && value.isNotEmpty) {
  result[key] = value;
}
```

The unused parser is at `dartastic_opentelemetry_api/lib/src/api/trace/trace_state.dart:28`. The re-injection path is at `w3c_trace_context_propagator.dart:117`.

## How to correct it

1. Parse with `TraceState.fromString`, or apply the same rules.
2. Check each member against the W3C key and value grammar.
3. Trim only the optional whitespace around a list member. Never trim inside a value.
4. Treat a duplicate key as invalid.
5. Stop at 32 members.
6. Drop an invalid member, so only a valid header is ever propagated.

NOTE: #109 covers unvalidated `TraceState` construction through the API. Both paths must validate.

## Notes

- Source finding files: `fable-ultra/context-04-tracestate-not-validated.md`, `fable-medium/propagation-tracestate-validation.md`
- Parent issue: #95

yuzurihaaa:
Closes #93. The SDK propagator half stays open in MindfulSoftwareLLC/dartastic_opentelemetry#195.

## The problem

W3C Trace Context allows one entry for each key. `TraceState.fromString` put the members into a map, so a repeated key was not an error. The last member replaced the first one, and the parser gave back a state that the caller sends on again.

OpenTelemetry Specification v1.60.0, [Trace API, TraceState](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#tracestate):

> `TraceState` MUST at all times be valid according to rules specified in [W3C Trace Context specification](https://www.w3.org/TR/trace-context/#tracestate-header-field-values).

"At all times" covers a state that comes from a header, not only a state that comes from a mutation.

## The change

One condition in the parse loop. `fromString` skips a member if the map already holds that key, so the first entry stays. The Dartdoc now says what the parser drops.

Nothing else changes. The key grammar, the value grammar, the limit of 32 members, `trim()` and `split('=')` all stay as they are.

## Tests

Six tests in `trace_state_spec_compliance_test.dart`: a repeated key, an invalid member with the valid members kept, a leading space inside a value, an empty list member, the limit of 32 members, and a round trip through `toString`.

`dart test` passes 233 tests. `dart analyze` reports no new problems, and no errors or warnings.

## Scope

Issue #93 lists six corrections. This pull request makes correction 4, "treat a duplicate key as invalid". Corrections 2, 5 and 6 are already on `main`. Correction 1 is for the [propagator](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/context/api-propagators.md#w3c-trace-context-requirements), which is the SDK half. Correction 3 is partly here, because the code never trims a value, so a leading space stays.

Three things stay out:

- The key grammar stays at W3C Trace Context Level 1, with the `tenant-id@system-id` format. The Trace API section links Level 1, and issue #49 tracks the multi-tenant format with the upstream project.
- `trim()` removes more whitespace than W3C counts as optional whitespace, for example a newline. This is not one of the four faults in #93.
- Issue #62 (`fromMap` accepts keys and values that nothing validates) and issue #61 (the order of entries after `put`) stay open.

Assisted-by: Claude Opus 5


lib/src/api/trace/trace_state.dart
```diff
@@ -36,7 +44,10 @@ class TraceState {
 
     for (var pair in pairs) {
       final keyValue = pair.trim().split('=');
+      // W3C allows one entry for each key. A repeated key makes the header
+      // invalid, so the first entry stays and the later ones are dropped.
```
michaelbushe: The code is good.  Please fix the comment.  It's not that the header is invalid, it just places it properly. "Only one entry per key is allowed because the entry represents that last position in the trace. Hence vendors must overwrite their entry upon reentry to their tracing system.

lib/src/api/trace/trace_state.dart
```diff
@@ -36,7 +44,10 @@ class TraceState {
...
       if (keyValue.length == 2 &&
+          !entries.containsKey(keyValue[0]) &&
           _isValidKey(keyValue[0]) &&
           _isValidValue(keyValue[1])) {
```
michaelbushe: Additionally, if the key or value is invalid, report it to OTelErrorHandling.

CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.4-wip]
 
+### Fixed
+
+- `TraceState.fromString` drops a list member if the key repeats. W3C Trace
+  Context allows one entry for each key, so a repeated key makes the header
```
michaelbushe: Same as above, not invalid but "because the entry represents that last position in the trace"

michaelbushe (changes requested):
Nice tests!  Some minor changes.

lib/src/api/trace/trace_state.dart
```diff
@@ -36,12 +49,20 @@ class TraceState {
...
       final keyValue = pair.trim().split('=');
-      if (keyValue.length == 2 &&
-          _isValidKey(keyValue[0]) &&
-          _isValidValue(keyValue[1])) {
-        entries[keyValue[0]] = keyValue[1];
-        if (entries.length >= _maxKeyValuePairs) break;
+      if (keyValue.length != 2 ||
+          !_isValidKey(keyValue[0]) ||
+          !_isValidValue(keyValue[1])) {
+        OTelErrorHandling.report(ArgumentError(
```
robert-northmind, line 62: Should we skip empty and whitespace members before reporting an error here? 
[W3C explicitly allows these](https://www.w3.org/TR/trace-context/#tracestate-header-field-values), like the empty part in vendor=v,,other=w.
Right now that reaches the error handler.

Could we keep those members silent and add a test checking that the handler isn't called? The current empty-member test only checks the returned entries, so it wouldn't catch this.

yuzurihaaa: Ouhhh.... good point. That should reduce the noise too.

robert-northmind (approved):

michaelbushe (approved):
The docs are **beautifully written**.  Just fix up the conflict and this is good to go.

PR #117, opened 2026-08-31
fix(api): drop a span event with an empty name instead of throwing

### LINKED ISSUE #69: [BUG] Trace API: addEvent throws ArgumentError on an empty event name
> **Migrated from MindfulSoftwareLLC/dartastic_opentelemetry#119** — the fix belongs entirely in `dartastic_opentelemetry_api`; the SDK issue is closed as moved.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Error handling, Basic error handling principles
**URL:** https://opentelemetry.io/docs/specs/otel/error-handling/#basic-error-handling-principles
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/error-handling.md#basic-error-handling-principles

>
> 1. API methods MUST NOT throw unhandled exceptions when used incorrectly by end users.
>    The API and SDK SHOULD provide safe defaults for missing or invalid arguments.

## The problem

`addEvent`, `addEventNow`, and `addEvents` throw an `ArgumentError` when the event name is empty. An end user who passes an empty name gets an unhandled exception at run time.

The current test suite asserts the wrong behavior. `span_test.dart:347` contains `test('addEvent throws if name is empty'...)`.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/trace/span.dart` (around line 287), and the test at `dartastic_opentelemetry_api/test/unit/api/trace/span_test.dart:347`.

## How to correct it

1. Remove the `throw ArgumentError(...)` from the event creation code in `Span` and in `SpanEvent`.
2. Drop the event instead, and write a warning:

```dart
OTelLog.warn('Event name cannot be empty, dropping event.');
```

3. Change the unit test. The test must assert that no exception occurs. The test can also assert that a warning is logged.

## Notes

- Found by: Gemini 3.1 Pro (run 1)
- Source finding files: `gemini-1/api-span-addevent-throws.md`
- Parent issue: #95

yuzurihaaa:
Closes #69.

## The problem

`SpanEventCreate.create` threw an `ArgumentError` when the event name was empty. The throw reached the user through `OTelAPI.spanEvent('')`, through `Span.addEventNow('')`, and through `Span.addEvents()` with an empty key.

OpenTelemetry Specification v1.60.0, [Error handling, Basic error handling principles](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/error-handling.md#basic-error-handling-principles):

> API methods MUST NOT throw unhandled exceptions when used incorrectly by end users. The API and SDK SHOULD provide safe defaults for missing or invalid arguments.

An empty name is incorrect use by the end user. Telemetry code must not stop the application.

## The change

The throw is removed from `SpanEventCreate.create`. A span now drops an event that has an empty name, and sends an `ArgumentError` to the error handler. [#102](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/102) made the handler the route for a suppressed error, and `error_report_routing_test.dart` (api#94) keeps a dropped value away from a plain `OTelLog.warn` line.

All three event methods go through one private test, `APISpan._dropsEmptyName`:

- `addEvent` tests the name of the event that the caller gives.
- `addEventNow` tests the name before it makes the event, so a dropped event is never allocated.
- `addEvents` tests each key, and drops only the entry that has the empty name.

The test is one `String.isNotEmpty` call, and it comes after the `_modifiable` test. A span that is ended, or a `NonRecordingSpan`, does no work and reports nothing.

The signatures do not change. `OTelAPI.spanEvent` and `OTelFactory.spanEvent` still give back a non-nullable `SpanEvent`, thus this is not a breaking change.

The issue text asks for `OTelLog.warn`. This pull request uses `OTelErrorHandling.report` instead, because api#94 made the handler the route for a dropped value, and the default handler still writes a log line.

## Tests

Two tests in `span_test.dart` take the place of `addEvent throws if name is empty`:

- All three event methods with an empty name make no exception, and the span holds no events.
- `addEvents` with one empty key and one good key keeps the good entry.

One test in `error_report_routing_test.dart` makes sure that the drop goes to the error handler and not to a warn-level log line.

`dart test` passes 1175 tests. `dart analyze` gives no errors and no new problems.

## Scope

Only the empty event name. The audit found other API methods that throw on incorrect input. Each one has its own issue, and this pull request does not touch them:

- #70 Meter validates instrument names and throws.
- #80 empty string and empty array attribute values are rejected and throw.
- #81 typed `Attributes` getters throw a `StateError` on a type mismatch.
- `OTelAPI.traceIdFrom` and `OTelAPI.spanIdFrom` throw a `FormatException` on a bad hex string. #91 covers the parse rules, but not the throw.

Argument errors in `OTelAPI.initialize` stay. They occur at initialization, and the specification permits a fast failure there.

Assisted-by: Claude Opus 5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   implementations need to be safe for concurrent use, which logs/api.md makes
   a MUST. Comments only, no behavior change
   ([#121](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/121)).
+- Span events with an empty name are dropped, not thrown. `OTelAPI.spanEvent('')`,
+  `addEvent`, `addEventNow` and `addEvents` no longer throw an `ArgumentError`;
+  the event is dropped and the error goes to the error handler, per
+  error-handling.md
+  ([#117](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/117)).
 
 ## [1.0.0-rc.3] - 2026-08-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   implementations need to be safe for concurrent use, which logs/api.md makes
   a MUST. Comments only, no behavior change
   ([#121](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/121)).
-- Span events with an empty name are dropped, not thrown. `OTelAPI.spanEvent('')`,
-  `addEvent`, `addEventNow` and `addEvents` no longer throw an `ArgumentError`;
-  the event is dropped and the error goes to the error handler, per
-  error-handling.md
+- Span events with an empty name are dropped and reported to `OTelErrorHandler`.
+  `OTelAPI.spanEvent('')`, `addEvent`, `addEventNow` and `addEvents` no longer
+  throw an `ArgumentError`, per error-handling.md
   ([#117](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/117)).
 
 ## [1.0.0-rc.3] - 2026-08-27
--- a/lib/src/api/trace/span.dart
+++ b/lib/src/api/trace/span.dart
@@ -4,6 +4,7 @@
 import 'package:meta/meta.dart';
 import '../../factory/otel_factory.dart';
 import '../../util/default_time_provider.dart';
+import '../../util/otel_error_handler.dart';
 import '../../util/time_provider.dart';
 import '../common/attributes.dart';
 import '../common/instrumentation_scope.dart';
@@ -277,7 +278,7 @@ class APISpan {
   /// "standard event names and keys" which have prescribed semantic meanings.
   /// See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/README.md
   void addEvent(SpanEvent spanEvent) {
-    if (_modifiable) {
+    if (_modifiable && !_dropsEmptyName(spanEvent.name)) {
       _spanEvents ??= [];
       _spanEvents!.add(spanEvent);
     }
@@ -288,7 +289,7 @@ class APISpan {
     if (OTelFactory.otelFactory == null) {
       throw StateError('Call initialize() first.');
     }
-    if (_modifiable) {
+    if (_modifiable && !_dropsEmptyName(name)) {
       _spanEvents ??= [];
       // Source the event timestamp from this span's TimeProvider so events
       // share the same clock as start/end. Bypasses the static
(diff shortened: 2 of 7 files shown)


lib/src/api/trace/span.dart
```diff
@@ -277,7 +278,7 @@ class APISpan {
   /// "standard event names and keys" which have prescribed semantic meanings.
   /// See https://github.com/open-telemetry/semantic-conventions/blob/main/docs/README.md
   void addEvent(SpanEvent spanEvent) {
-    if (_modifiable) {
+    if (_modifiable && !_dropsEmptyName(spanEvent.name)) {
```
michaelbushe: Please rename _dropsEmptyName to _hasEmptyName - it's the code after this that does the dropping.

lib/src/api/trace/span_event_create.dart
```diff
@@ -11,19 +11,18 @@ part of 'span_event.dart';
 class SpanEventCreate {
   /// Creates a new SpanEvent with the given parameters.
   ///
-  /// [name] The name of the event (required, must not be empty)
+  /// [name] The name of the event
   /// [timestamp] The time at which the event occurred
   /// [attributes] Optional attributes providing additional context
   ///
-  /// Throws an ArgumentError if name is empty.
+  /// This method does not throw. error-handling.md forbids a throw for
```
michaelbushe: You don't have to comment what we don't do. :)

lib/src/api/trace/span_event_create.dart
```diff
@@ -11,19 +11,18 @@ part of 'span_event.dart';
...
+  /// incorrect user input. A span drops an event that has an empty name.
+  /// See https://opentelemetry.io/docs/specs/otel/error-handling/#basic-error-handling-principles
   static SpanEvent create({
     required String name,
     required DateTime timestamp,
     Attributes? attributes,
   }) {
-    if (name.isEmpty) {
```
michaelbushe, line 24: We need to call the OTelErrorHandler here too.

test/unit/api/trace/span_test.dart
```diff
@@ -345,13 +345,26 @@ void main() {
...
     });
 
-    test('addEvent throws if name is empty', () {
-      // NOTE: becomes non-throwing under api#69 (#119), a separate issue.
+    test('an event with an empty name is dropped, not thrown (api#69)', () {
+      // error-handling.md: an API method must not throw when the user
+      // calls it incorrectly. Every event path drops the event instead.
+      // https://opentelemetry.io/docs/specs/otel/error-handling/#basic-error-handling-principles
       final span = tracer.startSpan('test');
-      expect(
```
michaelbushe, line 351: Test that OTelErrorHandler gets called.

CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   implementations need to be safe for concurrent use, which logs/api.md makes
   a MUST. Comments only, no behavior change
   ([#121](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/121)).
+- Span events with an empty name are dropped, not thrown. `OTelAPI.spanEvent('')`,
```
michaelbushe: ", not thrown" => "and reported to OTelErrorHandler." (This is more complete and spans events aren't thrown)

michaelbushe (changes requested):

lib/src/api/trace/span_event_create.dart
yuzurihaaa: Updated.

michaelbushe (approved):
Excellent.

PR #118, opened 2026-08-31
fix(api): Context-driven span creation and root span support (#83, #84)

### LINKED ISSUE #83: [BUG] Trace API: span creation accepts Span and SpanContext as parent (API surface)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#106** — fix the API first; the SDK issue stays open and tracks the SDK-side work, which depends on this landing in an API release.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Trace API, Span Creation
**URL:** https://opentelemetry.io/docs/specs/otel/trace/api/#span-creation
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#span-creation

> The parent `Context` or an indication that the new `Span` should be a root `Span`.
> The API MAY also have an option for implicitly using
> the current Context as parent as a default behavior.
> This API MUST NOT accept a `Span` or `SpanContext` as parent, only a full `Context`.

## The problem

Both span creation methods accept a `parentSpan` and a `spanContext` parameter. The code uses these to select the parent. They also take precedence over the `context` parameter.

A bare `Span` or `SpanContext` skips the parenting rules of `Context`. As a result, `IsRemote` handling is lost. The wrap of a remote `SpanContext` in a `NonRecordingSpan` is also lost.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/trace/tracer.dart:99`

```dart
APISpan startSpan(
  String name, {
  Context? context,
  SpanContext? spanContext,
  APISpan? parentSpan,
  ...
```

`dartastic_opentelemetry_api/lib/src/api/trace/tracer.dart:138`

```dart
APISpan createSpan({
  required String name,
  SpanContext? spanContext,
  APISpan? parentSpan,
  ...
```

`tracer.dart:153` resolves the parent from these parameters:

```dart
final effectiveParentSpan = parentSpan ?? contextSpan;
```

The doc comment advertises the forbidden precedence. It says "1. If [spanContext] is provided... 2. If [parentSpan] is provided, it will be used as the parent".

## How to correct it

Remove `spanContext` and `parentSpan` from `startSpan` and `createSpan`. Accept only a `Context` and a root indication:

```dart
APISpan startSpan(
  String name, {
  Context? context,        // defaults to Context.current
  bool root = false,       // explicit root-span indication
  SpanKind kind = SpanKind.internal,
  Attributes? attributes,
  List<SpanLink>? links,
  DateTime? startTime,
});
```

A user who holds a remote `SpanContext` must wrap it in a `NonRecordingSpan`. Then the user puts that span into a `Context` and passes the `Context` as the parent.

CAUTION: This change removes public parameters. It breaks callers. Add the `breaking-change` label if triage confirms the violation.

## Notes

- Found by: Fable 5 (both runs), Grok 4.5, Gemini 3.1 Pro
- Source finding files: `fable-ultra/trace-api-01-span-parent-not-context.md`, `fable-medium/trace-api-parent-must-be-context.md`, `grok/trace/007-parent-span-parameter.md`, `gemini-fail-mindful/trace_span_creation_parent_context.md`
- Parent issue: #95

### LINKED ISSUE #84: [BUG] Trace API: startSpan has no start timestamp parameter (API surface)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#111** — fix the API first; the SDK issue stays open and tracks the SDK-side work, which depends on this landing in an API release.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Trace API, Span Creation
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/trace/api/#span-creation
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#span-creation

> The API MUST accept the following parameters:
>
> - ...
> - `Start timestamp`, default to current time. This argument SHOULD only be set when span creation time has already passed.

## The problem

`startSpan` has no `startTime` parameter. Only `createSpan` accepts one.

`startSpan` is the primary span creation API. A caller that creates a span after the real start time cannot record that time.

## Offending code

- `dartastic_opentelemetry_api/lib/src/api/trace/tracer.dart:99-118`
- `dartastic_opentelemetry/lib/src/trace/tracer.dart:243-252`

## How to correct it

Add an optional `DateTime? startTime` parameter to `startSpan`. Pass the value through the sampling path and into the span constructor.

## Notes

- Found by: Grok 4.5
- Source finding files: `grok/trace/008-startspan-missing-start-time.md`
- Parent issue: #95

harshitt13:
# Description

Implements the API side of [dartastic_opentelemetry#106](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry/issues/106): span creation follows the OpenTelemetry spec's Span Creation rules, taking the parent exclusively from `Context`.

### Key changes

- **BREAKING**: removed the `parentSpan` and `spanContext` parameters from `APITracer.startSpan` and `APITracer.createSpan`. Parenting is now driven by the `Context` (or `Context.current`).
- **Added**: `root: true` on `startSpan`/`createSpan` forces a root span regardless of context.
- **Added**: optional `startTime` on `startSpan`.
- **Precedence**: `root` > remote `SpanContext` > local `Span` > valid non-remote `SpanContext` > new root, resolved by a single helper shared by the with-SDK and no-SDK paths, so a given `Context` parents identically either way. A candidate whose `SpanContext` is invalid is skipped, so an invalid parent yields a root span rather than an `ArgumentError`.
- When a remote `SpanContext` identifies a `Span` in the same `Context`, that `Span` is kept as the new span's `parentSpan` (thanks @robert-northmind). A `Span` from a different trace is still ignored, since the remote context wins.

### Migration

- `parentSpan: parent` becomes `context: Context.current.withSpan(parent)`.
- `spanContext: sc` has no direct equivalent. It assigned `sc` to the new span verbatim, span ID included, which the spec does not provide for. Putting `sc` on the `Context` makes the new span a *child* of it: new span ID, `parentSpanId` set to `sc.spanId`. To get a span carrying `sc` unchanged, use `OTelAPI.nonRecordingSpan(sc)`.

### Resolved: non-remote `SpanContext` on a `Context`

I originally raised this as an open question. Decision: a valid non-remote `SpanContext` placed on a `Context` with no `Span` **is** honored as a parent, rather than silently falling back to a new root. The spec's Span Creation section does not cover this combination, but the no-SDK path already honored it, `withSpanContext`/`copyWithSpanContext` are public API, and silently starting a new trace is the worst failure mode available: a broken trace with no error. Behavior is documented on `startSpan` and covered by tests.

### Note for reviewers

The SDK's `Tracer._startSpanInternal` currently calls `_delegate.createSpan(spanContext:, parentSpan:)`, which are the parameters this PR removes. That call site needs either a paired SDK change or an `@internal` seam on the API side before this can land. Happy to follow whichever the code owner prefers.

### Fixes

- Fixes #83
- Fixes #84


lib/src/api/trace/tracer.dart
```diff
@@ -168,70 +189,66 @@ class APITracer {
...
         effectiveSpanContext = OTelFactory.otelFactory!.spanContext(
           traceId: contextSpanContext.traceId,
           spanId: OTelFactory.otelFactory!.spanId(),
           parentSpanId: contextSpanContext.spanId,
           traceFlags: contextSpanContext.traceFlags,
           traceState: contextSpanContext.traceState,
         );
-      } else if (effectiveParentSpan != null) {
-        // Create child context from parent
+        effectiveParentSpan = null;
```
robert-northmind: Should we preserve the parent span object here when it contains the same remote context we are using?

For example, we can put OTelAPI.nonRecordingSpan(remote) into the Context and create a child from it. Before this change, the child's parentSpan returned that object and parentSpanContext returned remote. Now both will return null i think, even though the parent object is still available in contextSpan.

We should still ignore contextSpan when it represents a different parent. Could we handle these two cases separately? And extend the existing remote-parent test to check both getters? It currently only checks the IDs, which remain correct.

harshitt13: Good Catch! agreed and pushed the chnages in the latest commit

test/unit/api/trace/tracer_test.dart
```diff
@@ -77,14 +77,312 @@ void main() {
...
+    test('createSpan applies an explicit startTime', () {
+      final tracer = OTelAPI.tracer('test-tracer');
+      final startTime = DateTime.now().subtract(const Duration(minutes: 42));
+
+      final span = tracer.createSpan(name: 'test-span', startTime: startTime);
+
+      expect(span.startTime, equals(startTime));
+    });
+
+    test('startSpan forwards root and startTime through to createSpan', () {
```
michaelbushe, line 407: One case missing, and it is a MUST. `trace/api.md:664`: "It MUST still be possible to use an ended span as parent via a Context it is contained in." `_resolveParent` filters on `isValid`, not `isEnded`, so I think this already works, but nothing here proves it, and it is exactly the kind of guard someone adds later "to be safe".

```dart
test('an ended span in the context still parents a new span', () {
```

End the parent, put it in a context, start a child, assert `parentSpanContext` matches and the child is not a root.

CHANGELOG.md
```diff
@@ -7,8 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `APITracer.startSpan` and `APITracer.createSpan` now accept a `root: true`
+  parameter to force the creation of a root span, even when a parent span is
+  active in the context
+  ([#118](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/118)).
+- `APITracer.startSpan` now accepts an optional `startTime` parameter
+  ([#118](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/118)).
+
 ### Changed
 
+- **BREAKING**: `parentSpan` and `spanContext` parameters have been removed from
```
michaelbushe, line 51: Three things on this entry.

Maybe because of all the merges before it, this is a removal sitting under `### Added`. Keep a Changelog wants it under `### Removed`, with the `**BREAKING**:` prefix it already has. `root` and `startTime` stay under Added.

It is seventeen lines. The precedence list and the invalid-candidate rule are already in the `startSpan` dartdoc where they belong; the CHANGELOG needs the removal, the one-line migration for `parentSpan`, and the note that `spanContext: sc` has no equivalent. Four or five lines.  We need to keep things readable for users.

And one em dash in "span ID included — which".  Commas or periods are better.

lib/src/api/trace/tracer.dart
```diff
@@ -266,6 +273,75 @@ class APITracer {
...
+  /// not validate what is put into it, so a span with an invalid (all-zero)
+  /// SpanContext can sit in one; per trace/api.md an invalid parent means the
+  /// new span is a root, not an error.
+  ///
+  /// Static because it reads no per-tracer state: the parent depends only
+  /// on [context].
+  static ({SpanContext? parentSpanContext, APISpan? parentSpan}) _resolveParent(
+      Context context) {
+    final contextSpanContext = context.spanContext;
+    final contextSpan = context.span;
```
michaelbushe, line 331: Not a change request, a note for the doc. The spec has one lookup: the `Span` in the `Context`, and a remote `SpanContext` only gets into a `Context` by being wrapped in a span (`trace/api.md:430-437`). The five tiers exist because our `Context` carries a bare `spanContext` slot alongside `span`, which predates this PR. Your resolution of the two slots is sensible and it is good that it runs identically with and without an SDK. Worth one sentence in the `_resolveParent` doc saying the tiers come from our two slots rather than from the spec, so nobody later goes looking for them in api.md.

michaelbushe (dismissed):
Minor changes.

michaelbushe (approved):
Nice! Thank you.

PR #119, opened 2026-08-31
fix: validate TraceState keys/values during construction

### LINKED ISSUE #62: [BUG] Trace API: TraceState can be built with unvalidated keys and values
> **Migrated from MindfulSoftwareLLC/dartastic_opentelemetry#109** — the fix belongs entirely in `dartastic_opentelemetry_api`; the SDK issue is closed as moved.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Trace API, TraceState
**Requirement level:** MUST / MUST NOT
**URL:** https://opentelemetry.io/docs/specs/otel/trace/api/#tracestate
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#tracestate

> `TraceState` MUST at all times be valid according to rules specified in [W3C Trace Context specification](https://www.w3.org/TR/trace-context/#tracestate-header-field-values).
> Every mutating operations MUST validate input parameters.
> If invalid value is passed the operation MUST NOT return `TraceState` containing invalid data
> and MUST follow the [general error handling guidelines](https://opentelemetry.io/docs/specs/otel/error-handling/).

## The problem

Only `put()` validates keys and values. The public construction paths accept any map. No validation occurs.

A `TraceState` can therefore hold keys and values that W3C forbids. Examples are uppercase keys, a `=` or a `,` inside a value, and keys longer than 256 characters. The entry count limit of 32 is applied, but the content is not.

`toString()` then writes a corrupt `tracestate` header. For example, `TraceState.fromMap({'BadKey': 'a,b=c'})` produces `BadKey=a,b=c`. This header can break a downstream W3C parser.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/trace/trace_state_create.dart:11`

```dart
static TraceState create(Map<String, String>? entries) {
  if (entries == null || entries.isEmpty) {
    return TraceState._({});
  }
  // Enforce 32 key-value pair limit
  ...
  return TraceState._(entries);   // no key/value validation
}
```

Public entry points that reach this code are `TraceState.fromMap(...)` (`trace_state.dart:51`) and the `OTelAPI` and `OTelFactory` `traceState(...)` factory methods. Each passes the user map through without a check.

## How to correct it

Apply the existing `_isValidKey` and `_isValidValue` checks in `TraceStateCreate.create`. Drop an invalid entry and write a warning. Do not store it.

```dart
static TraceState create(Map<String, String>? entries) {
  final valid = <String, String>{};
  entries?.forEach((k, v) {
    if (valid.length < 32 && TraceState._isValidKey(k) && TraceState._isValidValue(v)) {
      valid[k] = v;
    } else {
      OTelLog.warn('Invalid TraceState entry "$k" dropped.');
    }
  });
  return TraceState._(valid);
}
```

## Notes

- Source finding files: `fable-ultra/trace-api-04-tracestate-unvalidated-construction.md`
- Parent issue: #95

benjaben:
## Summary
- `TraceStateCreate.create` (the single choke point behind `TraceState.fromMap`, `TraceState.empty()`, and the `OTelAPI`/`OTelFactory` `traceState(...)` factories) only enforced the 32-entry limit; `put()` was the only mutator that validated keys/values against the W3C tracestate grammar.
- `TraceState.fromMap({'BadKey': 'a,b=c'})` could therefore produce a `TraceState` that serializes to a corrupt `tracestate` header.
- `create` now validates each entry with the existing `TraceState._isValidKey`/`_isValidValue` checks, drops invalid entries, and reports them via `OTelErrorHandling.report` (same path `put()` uses), while still capping at 32 entries.

Fixes #62.

## Test plan
- [x] `dart analyze` — clean
- [x] `dart test` — full suite passes (1176 tests), 100% coverage maintained
- [x] Added `TraceState construction drops invalid entries` test group covering invalid key, invalid value, and full round-trip
- [x] Fixed a pre-existing test (`span_context_test.dart`) that unknowingly relied on the bug via an uppercase key (`'newVendor'`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed (spec compliance)
 
+- `TraceState` construction (`fromMap`, `OTelAPI`/`OTelFactory` `traceState(...)`)
+  now validates keys and values against the W3C tracestate grammar, dropping
+  invalid entries instead of silently accepting them
+  ([#119](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/119)).
 - **BREAKING**: `IdGenerator.hexToBytes`, and so `OTelAPI.traceIdFrom` and
   `OTelAPI.spanIdFrom`, no longer accept anything outside lowercase hex
   ([#112](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/112)).
--- a/lib/src/api/trace/trace_state_create.dart
+++ b/lib/src/api/trace/trace_state_create.dart
@@ -6,29 +6,24 @@ part of 'trace_state.dart';
 /// Internal constructor access for TraceState
 @internal
 class TraceStateCreate {
-  /// Creates a TraceState, only accessible within library
-  /// Enforces 32 key-value pair limit per W3C spec
+  /// Drops invalid entries; keeps at most 32.
   static TraceState create(Map<String, String>? entries) {
     if (entries == null || entries.isEmpty) {
       return TraceState._({});
     }
 
-    // Enforce 32 key-value pair limit
-    if (entries.length > 32) {
-      // Take only the first 32 entries
-      final limitedEntries = <String, String>{};
-      var count = 0;
-
-      for (final entry in entries.entries) {
-        if (count >= 32) break;
-
-        limitedEntries[entry.key] = entry.value;
-        count++;
+    final validEntries = <String, String>{};
+    for (final entry in entries.entries) {
+      if (validEntries.length >= 32) break;
+      if (TraceState._isValidKey(entry.key) &&
+          TraceState._isValidValue(entry.value)) {
+        validEntries[entry.key] = entry.value;
+      } else {
+        OTelErrorHandling.report(ArgumentError(
+            'Invalid TraceState key or value: "${entry.key}"; entry dropped.'));
       }
-
-      return TraceState._(limitedEntries);
     }
 
-    return TraceState._(entries);
+    return TraceState._(validEntries);
   }
 }
--- a/lib/src/api/trace/trace_state_create.dart
+++ b/lib/src/api/trace/trace_state_create.dart
@@ -6,7 +6,10 @@ part of 'trace_state.dart';
 /// Internal constructor access for TraceState
 @internal
 class TraceStateCreate {
-  /// Drops invalid entries; keeps at most 32.
+  /// Creates a TraceState, only accessible within library.
+  /// Enforces the W3C 32 key-value pair limit. Entries with an invalid key
+  /// or value are dropped and reported via [OTelErrorHandling.report]
+  /// rather than throwing, per trace/api.md.
   static TraceState create(Map<String, String>? entries) {
     if (entries == null || entries.isEmpty) {
       return TraceState._({});
--- a/test/unit/api/trace/trace_state_spec_compliance_test.dart
+++ b/test/unit/api/trace/trace_state_spec_compliance_test.dart
@@ -47,6 +47,24 @@ void main() {
     });
   });
 
+  group('TraceState construction drops invalid entries', () {
+    test('fromMap drops an entry with an invalid key', () {
+      final result = TraceState.fromMap({'BadKey': 'a', 'goodkey': 'b'});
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.4-wip]
 
+### Fixed
+
+- `TraceState` construction (`fromMap`, `OTelAPI`/`OTelFactory` `traceState(...)`)
+  now validates keys and values against the W3C tracestate grammar, dropping
+  invalid entries instead of silently accepting them.
```
yuzurihaaa: nit: PR link here

yuzurihaaa (approved):
LGTM. A small nit comment on CHANGELOG based on discussion on Slack.

lib/src/api/trace/trace_state_create.dart
```diff
@@ -6,29 +6,24 @@ part of 'trace_state.dart';
...
-      var count = 0;
-
-      for (final entry in entries.entries) {
-        if (count >= 32) break;
-
-        limitedEntries[entry.key] = entry.value;
-        count++;
+    final validEntries = <String, String>{};
+    for (final entry in entries.entries) {
+      if (validEntries.length >= 32) break;
```
michaelbushe, line 20: You rewrote the 32-entry cap here and the only test covering it is `trace_state_test.dart:79`, which asserts `entries.length <= 32`. That passes at zero, so if a valid key ever stopped validating, this loop could drop everything and the suite would stay green. Tighten it to `equals(32)` while you're in here.

test/unit/api/trace/trace_state_spec_compliance_test.dart
```diff
@@ -47,6 +47,24 @@ void main() {
...
+      final result = TraceState.fromMap({'vendor': 'a,b=c', 'goodkey': 'b'});
+      expect(result.entries, equals({'goodkey': 'b'}));
+    });
+
+    test('fromMap never produces a TraceState containing invalid data', () {
+      final result = TraceState.fromMap({'BadKey': 'a,b=c'});
+      expect(result.entries, isEmpty);
+      expect(result.toString(), isEmpty);
+    });
+  });
```
michaelbushe, line 87: Three cases missing from this group.

You added the `OTelErrorHandling.report` call and nothing asserts it fires. Install a handler, assert one `ArgumentError` arrives, and `addTearDown(OTelErrorHandling.resetToDefault)`.

The length limits are untested too: a key over 256 characters and a value over 256 characters.

lib/src/api/trace/trace_state_create.dart
```diff
@@ -6,29 +6,24 @@ part of 'trace_state.dart';
 /// Internal constructor access for TraceState
 @internal
 class TraceStateCreate {
-  /// Creates a TraceState, only accessible within library
-  /// Enforces 32 key-value pair limit per W3C spec
+  /// Drops invalid entries; keeps at most 32.
```
michaelbushe: This drops more than it adds. The old comment said where 32 comes from, and now nothing does. Keep a line noting it's the W3C limit, and say that invalid entries are dropped and reported rather than throwing, since that's the part of `trace/api.md` this is implementing.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.4-wip]
 
+### Fixed
```
michaelbushe: #121 landed a `### Fixed (spec compliance)` heading under rc.4-wip. Fold into that on the rebase instead of adding a second `### Fixed`.

michaelbushe (changes requested):

CHANGELOG.md
benjaben: Updated!

lib/src/api/trace/trace_state_create.dart
benjaben: Updated!

CHANGELOG.md
benjaben: Updated!

robert-northmind (approved):

michaelbushe (approved):
Nice changes, good test coverage.

PR #123, opened 2026-09-01
[API] feat: implement AnyValue attribute hierarchy and strict compliance (#95)

### LINKED ISSUE #95: [BUG] Common: AnyValue is incomplete — no maps, nested arrays, bytes, or null (API)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#236** — fix the API first; the SDK issue stays open and tracks the SDK-side work, which depends on this landing in an API release.

## Spec requirement

**Requirement level:** MUST

> - The attribute value MUST be one of types defined in [AnyValue](https://opentelemetry.io/docs/specs/otel/common/#anyvalue).

`AnyValue` covers a map, a nested array, a byte array, and an empty or null value.

## The problem

The attribute value supports a primitive and a list of one primitive type only.

A map is not supported. A nested array is not supported. A byte array is not supported. An empty or null value is not supported.

Two consequences follow. A complex attribute cannot be represented. A nested log body cannot be represented either.

## Offending code

- `dartastic_opentelemetry_api/lib/src/api/common/attribute.dart`
- `dartastic_opentelemetry_api/lib/src/api/factory/otel_api_factory.dart:170-216`

## How to correct it

Add a full `AnyValue` type. Use it for an attribute value and for a log body.

NOTE: The empty-value clause of `AnyValue` is a separate issue in this same section. That one is smaller and can land first.

## Notes

- Found by: Grok 4.5
- Source finding files: `grok/common/004-anyvalue-incomplete.md`
- Parent issue: #95

harshitt13:
Closes #95

## Description

This PR brings `AnyValue` and the attribute representation into line with the OpenTelemetry specification. Previously the API had no support for maps, nested arrays, bytes or null, and had gaps in how attributes were represented and unwrapped internally.

### Core Changes

1. **`AnyValue` hierarchy**: a sealed hierarchy (`AnyValueString`, `AnyValueBool`, `AnyValueInt`, `AnyValueDouble`, `AnyValueArray`, `AnyValueMap`, `AnyValueBytes`, `AnyValueNull`) covering every attribute and log body type in the specification.
2. **`Attribute` refactor**: the generic `Attribute<T>` becomes a concrete `Attribute` carrying an `AnyValue` payload.
3. **`Attributes` unboxing**: `_getTyped<T>` unboxes `AnyValue` subtypes, promoting ints to doubles when a double is asked for.
4. **Attribute values accept the whole model**: common.md's Attribute section says the value MUST be one of the types defined in `AnyValue`, so a map, a nested array, a byte array and null are all valid attribute values, not just log bodies. That is the gap #95 was filed about.

### Breaking changes

- **`Attribute<T>` is no longer generic.** Every explicit type argument has to go: `Attribute<String> a = ...` becomes `Attribute a = ...`, and `List<Attribute<Object>>` becomes `List<Attribute>`. `attribute.value` is now an `AnyValue` rather than the raw Dart value, so read it through the typed `Attributes` getters or unwrap it. Construction is unaffected, since the `Attribute` constructor was already private.
- **`LogRecord.body` is an `AnyValue?`** rather than an `Object?`, so a reader gets the typed model. `APILogger.emit(body: ...)` still takes a plain `Object?`, so callers are unaffected, and it accepts an `AnyValue` too, so `emit(body: record.body)` forwards a record without re-wrapping it. An SDK boxes the body with the new `APILogger.bodyToAnyValue` static, which reports a body it cannot convert through `OTelErrorHandling` and drops it rather than throwing into the caller's logging path.
- **A `Uint8List` attribute value is stored as a byte array** rather than flattened into an integer list. It used to match the old `value is List<int>` branch, so `getIntList` returned the bytes as ints. It is now an `AnyValueBytes`, read back through `attributes.toMap()[key]!.value`. `Int32List` and other typed lists are unaffected and still convert as arrays.
- **The map entry points take `Map<String, Object?>`.** `Attributes.of`, `OTelAPI.attributesFromMap`, `OTelFactory.attributesFromMap` and the `toAttributes()` extension widened their parameter so a null attribute value can be passed. An `OTelFactory` implementation overriding `attributesFromMap` must widen to match.
- **`AnyValueArray`, `AnyValueMap` and `AnyValueBytes` are no longer `const` constructible.** They copy their contents into unmodifiable collections, which a `const` constructor cannot do. The five scalar subtypes remain `const`.
- **An empty attribute key is dropped and reported** rather than stored. The factories do not throw: error-handling.md makes throwing at end-user misuse a MUST NOT, so the drop happens in `Attributes._` and is reported.

### Correctness fixes found

- `Uint8List` now converts to `AnyValueBytes`. It implements `List<int>`, so the list branch caught it first and `AnyValueBytes` was unreachable from `fromObject`. Other typed lists such as `Int32List` still convert as arrays.
- `AnyValueBytes` rejects elements outside 0-255 rather than masking them into range the way a bare `Uint8List` would. Silently truncating a byte corrupts an opaque payload. A `Uint8List` input skips the scan, being in range by construction.
- Conversion is depth limited to 32 levels in both directions. A deeper structure throws an `ArgumentError` that `attrsFromMap` catches, dropping and reporting that attribute.
- The collection subtypes copy defensively into unmodifiable collections. They are value-equal and used as keys, so sharing the caller's list would let a later mutation change a stored value's `hashCode`. Their deep hashes are cached, since the values are immutable.
- `Attribute.toString` renders through `AnyValue.toString` rather than the raw payload, which previously printed an array as `AttributeValue([Instance of 'AnyValueString', ...])`. The rendering is bounded in both directions, because `toString` runs where an exception or a megabyte of output is least welcome: nesting past the depth limit truncates rather than throwing, and bytes render as `<N bytes>` rather than their contents. A String nested in a collection is quoted, so `["a", "b"]` and `["a, b"]` are distinguishable.
- `getDouble` promotes a stored int, and `getDoubleList` promotes an all-int array. Promotion is one way: `getInt` on a stored double still returns null. This also absorbs the web's single number type, where `2.0 is int` is true and a whole-valued double is stored as an `AnyValueInt`.

### Conversion of arbitrary values

`AnyValue.fromObject` follows "Mapping Arbitrary Data to OTLP AnyValue". A type with no dedicated mapping is converted through its `toString()`, per that document's Other Values rules. If `toString()` itself throws, the failure is reported through `OTelErrorHandling` and the value becomes an empty `AnyValue`, which is the last resort the same section prescribes. The intermediate `bytes_value` step is skipped, since Dart offers no general way to serialise an arbitrary object to bytes.

Two conditions are still refused outright, and both throw an `ArgumentError` that `attrsFromMap` and `fromJson` catch, drop and report: a non-String map key, and a structure nested more than 32 levels deep. The mapping doc says non-String keys MUST be stringified too; that is a separate rule with its own test surface and I am happy to do it here or as a follow-up.

Worth knowing: a class with no custom `toString()` stringifies to `Instance of 'MyClass'`, which is rarely the telemetry anyone wanted. That is the spec-prescribed outcome, so it is documented on `fromObject` rather than detected.

### Not changed, deliberately

`AnyValue.toJson` is `unwrap`, so plain Dart objects, bytes included. It is **not** the OTLP/JSON wire encoding, which is tagged (`{"intValue": "1"}`, with int64 as a String) and is produced SDK-side from the protobuf model. An earlier revision of this PR base64-encoded bytes in `toJson` and claimed OTLP/JSON; a reviewer correctly pointed out that this was a near-duplicate of `unwrap` with a surprising difference and a false claim in the docs, so it was reverted rather than extended. If you want a real tagged encoder in the API, that is a separate PR and I am happy to open it.

The typed `Attributes` getters still return only the primitive and homogeneous-list shapes. `getString` on a map attribute returns null and reports a type mismatch, the same as asking for the wrong primitive type. Reach the rest through `attributes.toMap()[key]!.value`, which hands back the `AnyValue`. A `getAnyValue(key)` convenience would be easy to add if it is wanted.

## Downstream impact on the SDK

**The SDK does not compile against this branch.** Verified by pointing `dartastic_opentelemetry` at this branch through `dependency_overrides`: 15 analyzer errors and 184 test files failing to load. Eight of the errors are `wrong_number_of_type_arguments` on `Attribute<String>` and friends in `lib/src/otel.dart`, one in `exemplar.dart`, one `invalid_override` on `SDKLogRecord.body`, and five in `otel_sdk_test.dart`.

**The analyzer does not catch the more serious problem.** `log_record_transformer.dart` and `metric_transformer.dart` test attribute values with `is String` / `is List<String>` chains that now all evaluate false and fall through to an `else` branch doing `anyValue.stringValue = attr.value.toString()`. No compile error, no crash: every attribute would export as the stringification of an `AnyValue` wrapper object rather than the value inside it.

**No published SDK breaks.** The SDK pins `dartastic_opentelemetry_api: '>=1.0.0-rc.3 <1.0.0-rc.4'`, which excludes rc.4, so publishing does not break an existing user. The breakage lands when that constraint widens.

**Proposed sequencing:** I open a companion SDK PR rewriting both transformers to switch on the sealed hierarchy rather than patching the `is` chains, which makes the accidental stringification unreachable by construction and adds the `AnyValueMap`, `AnyValueArray` and `AnyValueBytes` cases setting the matching protobuf fields. Happy to land that first, land them together, or land this one first, whichever you prefer.

## Design notes

**Non-generic `Attribute`.** I weighed keeping `Attribute<T>` where `T extends AnyValue` against dropping the generic. Dropping it won: keeping `T` on the `Attributes` getters meant deep casting complexity, and an attribute is a heterogeneous key/value pair in the specification anyway.

**Where validation lives.** `Attributes._` is the single constructor every `Attributes` passes through: `attrsFromMap`, `fromJson`, `attributesFromList`, an `Attribute` passed straight through `attrsFromMap`, and the `copyWith*` methods. The non-empty-key rule is enforced there for that reason. An earlier revision of this PR also enforced a narrower *value* rule there, restricting attribute values to primitives and homogeneous arrays. A reviewer pointed out that this reimplemented the exact bug #95 was filed about, and they were right: common.md defines an attribute value as any `AnyValue` type. The value restriction is gone and only the key rule remains.

**A null inside an array is preserved.** common/README.md makes this a MUST where it cannot be prevented at compile time, and `['a', null, 'c']` now round-trips. The typed getters cannot express it, since `_getTyped` hands back `List<String>` rather than `List<String?>`, so read such an array through `toMap()[key]!.value` instead.

**The no-op `emit` does nothing at all.** An earlier revision converted the body so an unrepresentable one would be reported even with no SDK installed. A reviewer pointed out that logs/noop.md says the No-Op Logger accepts the parameters without validating them, and that the cost landed on exactly the users who installed no SDK. They were right, and there is a stronger argument still: with no SDK no record is produced, so there is nothing an unrepresentable body could corrupt, and reporting on the body alone while silently discarding every other field of the same record was arbitrary. `emit` is now empty, and a test using a read-counting `List` asserts it never touches the body.

**`emit(body:)` signature.** An earlier revision had `emit` take `AnyValue?`, which meant a caller passing a domain object got an `ArgumentError` thrown into their logging path. A reviewer flagged the ergonomic cost and I agree the error routing argument is stronger still: telemetry that fails should not take down the caller. `emit` now takes `Object?`, and an SDK boxes it through `APILogger.bodyToAnyValue`, reporting and dropping a body it cannot convert. `LogRecord.body` stays `AnyValue?` so the model reader gets the strong type. `bodyToAnyValue` passes an existing `AnyValue` straight through, so `emit(body: record.body)` forwards a record rather than re-wrapping it.

`bodyToAnyValue` is deliberately `static` rather than an instance method. `OTelLogger` in the SDK implements `APILogger` rather than extending it, and `implements` requires a concrete implementation of every public instance member. `@protected` is an annotation, not a visibility modifier, so a `@protected` instance method would have forced the SDK to write its own copy of the report-and-drop logic. As a static it stays out of the interface and the SDK calls `APILogger.bodyToAnyValue(body)` to get the identical behavior. Verified against the SDK's exact class shape.

## Authorship

- Assisted-by: Claude

### DIFF
--- a/lib/src/api/common/any_value.dart
+++ b/lib/src/api/common/any_value.dart
@@ -1,0 +1,203 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+import 'package:collection/collection.dart';
+import 'package:meta/meta.dart';
+
+/// Represents a value of any type supported by the OpenTelemetry specification.
+@immutable
+sealed class AnyValue {
+  const AnyValue();
+
+  /// Returns the underlying Dart object (not recursively unwrapped).
+  Object? get value;
+
+  Object? unwrap() {
+    return switch (this) {
+      AnyValueNull() => null,
+      AnyValueString(value: final v) => v,
+      AnyValueBool(value: final v) => v,
+      AnyValueInt(value: final v) => v,
+      AnyValueDouble(value: final v) => v,
+      AnyValueArray(value: final v) => v.map((e) => e.unwrap()).toList(),
+      AnyValueMap(value: final v) =>
+        v.map((k, val) => MapEntry(k, val.unwrap())),
+      AnyValueBytes(value: final v) => v,
+    };
+  }
+
+  /// Implements JSON serialization.
+  Object? toJson() => unwrap();
+
+  /// Creates an AnyValue from a String.
+  factory AnyValue.fromString(String value) = AnyValueString;
+
+  /// Creates an AnyValue from a boolean.
+  factory AnyValue.fromBool(bool value) = AnyValueBool;
+
+  /// Creates an AnyValue from an integer.
+  factory AnyValue.fromInt(int value) = AnyValueInt;
+
+  /// Creates an AnyValue from a double.
+  factory AnyValue.fromDouble(double value) = AnyValueDouble;
+
+  /// Creates an AnyValue from a List of AnyValues.
+  factory AnyValue.fromList(List<AnyValue> value) = AnyValueArray;
+
+  /// Creates an AnyValue from a Map of String to AnyValue.
+  factory AnyValue.fromMap(Map<String, AnyValue> value) = AnyValueMap;
+
+  /// Creates an AnyValue from a byte array.
+  factory AnyValue.fromBytes(List<int> value) = AnyValueBytes;
+
+  /// Creates an empty or null AnyValue.
+  factory AnyValue.empty() = AnyValueNull;
+
+  /// Creates an AnyValue by recursively converting a standard Dart object.
+  ///
+  /// Throws [ArgumentError] if it encounters an unsupported type.
+  factory AnyValue.fromObject(dynamic obj) {
+    if (obj == null) {
+      return const AnyValueNull();
+    } else if (obj is String) {
+      return AnyValueString(obj);
+    } else if (obj is bool) {
+      return AnyValueBool(obj);
+    } else if (obj is int) {
+      return AnyValueInt(obj);
+    } else if (obj is double) {
+      return AnyValueDouble(obj);
+    } else if (obj is List) {
+      return AnyValueArray(obj.map(AnyValue.fromObject).toList());
+    } else if (obj is Map) {
+      final map = <String, AnyValue>{};
+      obj.forEach((key, val) {
+        if (key is! String) {
+          throw ArgumentError(
+              'AnyValue map keys must be Strings, got ${key.runtimeType}');
+        }
+        map[key] = AnyValue.fromObject(val);
+      });
+      return AnyValueMap(map);
+    } else if (obj is DateTime) {
+      return AnyValueString(obj.toUtc().toIso8601String());
+    } else {
+      throw ArgumentError(
+          'Unsupported type in AnyValue conversion: ${obj.runtimeType}');
+    }
+  }
+}
+
+class AnyValueNull extends AnyValue {
+  const AnyValueNull();
+
+  @override
+  Object? get value => null;
+
+  @override
+  int get hashCode => null.hashCode;
+
+  @override
+  bool operator ==(Object other) => other is AnyValueNull;
+}
+
+class AnyValueString extends AnyValue {
+  @override
+  final String value;
+
+  const AnyValueString(this.value);
+
+  @override
+  int get hashCode => value.hashCode;
+
+  @override
+  bool operator ==(Object other) =>
+      other is AnyValueString && value == other.value;
+}
+
+class AnyValueBool extends AnyValue {
+  @override
+  final bool value;
+
+  const AnyValueBool(this.value);
+
+  @override
+  int get hashCode => value.hashCode;
+
+  @override
+  bool operator ==(Object other) =>
+      other is AnyValueBool && value == other.value;
+}
+
+class AnyValueInt extends AnyValue {
+  @override
+  final int value;
+
+  const AnyValueInt(this.value);
+
+  @override
+  int get hashCode => value.hashCode;
+
+  @override
+  bool operator ==(Object other) =>
+      other is AnyValueInt && value == other.value;
+}
+
+class AnyValueDouble extends AnyValue {
+  @override
+  final double value;
+
+  const AnyValueDouble(this.value);
+
+  @override
+  int get hashCode => value.hashCode;
+
+  @override
+  bool operator ==(Object other) =>
+      other is AnyValueDouble && value == other.value;
+}
+
+class AnyValueArray extends AnyValue {
+  @override
+  final List<AnyValue> value;
+
+  const AnyValueArray(this.value);
+
+  @override
+  int get hashCode => const DeepCollectionEquality().hash(value);
+
+  @override
+  bool operator ==(Object other) =>
+      other is AnyValueArray &&
+      const DeepCollectionEquality().equals(value, other.value);
+}
+
+class AnyValueMap extends AnyValue {
+  @override
+  final Map<String, AnyValue> value;
+
+  const AnyValueMap(this.value);
+
+  @override
+  int get hashCode => const DeepCollectionEquality().hash(value);
+
+  @override
+  bool operator ==(Object other) =>
+      other is AnyValueMap &&
+      const DeepCollectionEquality().equals(value, other.value);
+}
+
+class AnyValueBytes extends AnyValue {
+  @override
+  final List<int> value;
+
+  const AnyValueBytes(this.value);
+
+  @override
+  int get hashCode => const DeepCollectionEquality().hash(value);
+
+  @override
+  bool operator ==(Object other) =>
+      other is AnyValueBytes &&
+      const DeepCollectionEquality().equals(value, other.value);
+}
--- a/lib/src/api/common/attribute.dart
+++ b/lib/src/api/common/attribute.dart
@@ -1,74 +1,49 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-import 'package:collection/collection.dart';
 import 'package:meta/meta.dart';
+
+import 'any_value.dart';
 
 part 'attribute_create.dart';
 
 /// Represents a value for an attribute, associated with an attribute key.
 @immutable
-class Attribute<T extends Object> {
-  /// The DeepCollectionEquality object used for comparing list values.
-  /// This enables proper element-by-element comparisons for lists.
-  static final DeepCollectionEquality _listEquality =
-      const DeepCollectionEquality();
-
+class Attribute {
   /// The key (name) of this attribute.
   final String _key;
 
   /// The value of this attribute.
-  final T _value;
+  final AnyValue _value;
 
-  Attribute._(String key, T value)
+  Attribute._(String key, AnyValue value)
       : _key = key,
-        _value = value {
-    if (_value is String && (_value as String).isEmpty) {
-      throw ArgumentError('Attribute _value must not be an empty string');
-    }
-    if (_value is List) {
-      final valueList = _value as List;
-      if (valueList.isEmpty) {
-        throw ArgumentError('Attribute _value list must not be empty');
-      }
-      // No null-element guard: attributes are only created with
-      // non-nullable element types (List<String>, List<bool>, List<int>,
-      // List<double>), so sound null safety already rules nulls out.
-    }
-  }
+        _value = value;
 
   /// Gets the key (name) of this attribute.
   String get key => _key;
 
   /// Gets the value of this attribute.
-  T get value => _value;
+  AnyValue get value => _value;
 
   @override
   String toString() {
-    return 'AttributeValue($_value)';
+    return 'AttributeValue(${_value.value})';
   }
 
   @override
   bool operator ==(Object other) {
     if (identical(this, other)) return true;
 
-    if (other is! Attribute<T>) return false;
-
-    // Use deep equality for lists, otherwise default equality
-    final valuesAreEqual = value is List
-        ? _listEquality.equals(value, other.value)
-        : value == other.value;
+    if (other is! Attribute) return false;
 
     return runtimeType == other.runtimeType &&
         key == other.key &&
-        valuesAreEqual;
+        value == other.value;
   }
 
   @override
   int get hashCode {
-    // Use deep hash for lists, otherwise default hash
-    final valueHash =
-        value is List ? _listEquality.hash(value) : value.hashCode;
-    return Object.hash(key, valueHash);
+    return Object.hash(key, value);
   }
 }
--- a/lib/src/api/common/attribute_create.dart
+++ b/lib/src/api/common/attribute_create.dart
@@ -13,7 +13,7 @@ class AttributeCreate {
   /// @param name The name of the attribute
   /// @param value The value of the attribute
   /// @return A new Attribute instance
-  static Attribute<T> create<T extends Object>(String name, T value) {
+  static Attribute create(String name, AnyValue value) {
     return Attribute._(name, value);
   }
 }
--- a/lib/src/api/logs/logger.dart
+++ b/lib/src/api/logs/logger.dart
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import 'package:meta/meta.dart';
+import '../common/any_value.dart';
 import '../common/attributes.dart';
 import '../context/context.dart';
 import 'severity.dart';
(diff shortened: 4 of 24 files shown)


michaelbushe:
Update: I've been looking into this. It's a significant spec change so it's going to take me some more time.  I'm going to finish getting caught up on the smaller PRs and loop back to this..

lib/src/api/common/any_value.dart
```diff
@@ -0,0 +1,374 @@
...
+    } else if (obj is DateTime) {
+      // Timestamp.dateTimeToString, not toIso8601String: it is what
+      // attrsFromMap and Span.setDateTimeAttribute already emit, and it pins
+      // the fractional part to milliseconds instead of widening to
+      // microseconds whenever the DateTime happens to carry them.
+      return AnyValueString(Timestamp.dateTimeToString(obj));
+    } else {
+      throw ArgumentError(
+          'Unsupported type in AnyValue conversion: ${obj.runtimeType}');
+    }
```
robert-northmind: Should we maybe fall back to a `toString()` for unknown types here? 
We already convert DateTime into a string, and the [AnyValue mapping guidance](https://opentelemetry.io/docs/specs/otel/common/attribute-type-mapping/#other-values) says other values SHOULD be stringified when possible. **But note:** that guidance is still marked as **Development**, but maybe still a guarded fallback would fit better than dropping these values entirely.

Also, one concern about throwing here? If i read the code correctly, then the callers in attrsFromMap and bodyToAnyValue do catch this, but I think the AnyValue.fromObject(Object()) method is public right, and in that case we would not handle it. 

Could we try string conversion, then report errors through OTelErrorHandling and return an AnyValueNull if conversion fails? `toString()` itself can throw, so we'd need to handle that too. What do you think?

lib/src/api/common/attribute.dart
```diff
@@ -1,61 +1,59 @@
...
 
   /// Gets the value of this attribute.
-  T get value => _value;
+  AnyValue get value => _value;
 
   @override
   String toString() {
+    // AnyValue.toString, not `_value.value`, which renders an array as
+    // `[Instance of 'AnyValueString', ...]`. It also truncates rather than
+    // throwing on a deeply nested value.
```
robert-northmind: Maybe it is only me not understanding the comment here.
But i am not sure i fully get what you mean here, the part about the `_value.value`.
So here we call: `'AttributeValue($_value)'`. And that under the hood calls the `toString()` on the AnyValue right?

lib/src/api/common/attribute_create.dart
```diff
@@ -13,7 +13,97 @@ class AttributeCreate {
...
+  static bool isValidAttributeValue(AnyValue value) {
+    // A switch over the sealed hierarchy, not an if-chain, so a new subtype
+    // is a compile error here instead of silently becoming legal.
+    return switch (value) {
+      AnyValueString() ||
+      AnyValueBool() ||
+      AnyValueInt() ||
+      AnyValueDouble() =>
+        true,
+      AnyValueMap() || AnyValueBytes() || AnyValueNull() => false,
```
robert-northmind: Should we remove this restriction?
#95 specifically asks for maps, nested arrays, bytes and null as attribute values. The [v1.60 spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/common/README.md#attribute) includes these in `AnyValue`, alongside primitives like strings and booleans.
For example, `Attributes.of({'context': {'nested': true}})` currently drops the whole context attribute. Shouldn't we preserve it with the map as its value? We can still access it through `toMap()`, even without a dedicated typed getter.
Could we update the tests to check that these values survive creation and retrieval? We'd also need to allow nullable values through the map entrypoints.

lib/src/api/logs/logger.dart
```diff
@@ -59,17 +61,54 @@ class APILogger {
...
-    dynamic body,
+    Object? body,
     Attributes? attributes,
     String? eventName,
   }) {
-    // Base implementation is a no-op
+    // Base implementation is a no-op, but the body is still normalized so an
+    // unsupported value is reported here rather than by whichever SDK
+    // happens to be installed.
+    bodyToAnyValue(body);
```
robert-northmind: Should we keep this no-op `emit` empty and leave body conversion to the SDK implementation?
The [no-op spec](https://opentelemetry.io/docs/specs/otel/logs/noop/) says it must accept the parameters without validating them or logging errors. Currently an unsupported body reaches the error handler even without an SDK installed.
Checking `isEnabled()` avoids the call, but calling `emit` directly should maybe still do nothing.
We could keep the public static `bodyToAnyValue` helper for SDK reuse, and add tests that no-op emission neither reports errors nor traverses the body. What do you think?

lib/src/api/logs/logger.dart
```diff
@@ -59,17 +61,54 @@ class APILogger {
...
+  /// logging path. This mirrors what `attrsFromMap` does for attributes.
+  ///
+  /// Static, not an instance method: SDK loggers `implement` [APILogger] and
+  /// delegate rather than extending it, so an instance member would oblige
+  /// every one of them to supply its own implementation — the opposite of
+  /// sharing this one. Call it as `APILogger.bodyToAnyValue(body)`.
+  static AnyValue? bodyToAnyValue(Object? body) {
+    if (body == null) return null;
+    try {
+      return AnyValue.fromObject(body);
```
robert-northmind, line 114: Could we support both plain Dart values and already-wrapped `AnyValue` values here?
`LogRecord.body` is now an `AnyValue?`, so forwarding it through `emit(body: record.body)` seems like a natural thing to do. But if the SDK uses this helper, the wrapped value gets reported as unsupported and dropped.
We could return it directly before calling `fromObject`:

```dart
if (body is AnyValue) return body;
```

Then plain values still get converted, and existing `AnyValue` instances pass through unchanged. If we add this, then maybe we add tests for both paths too?

robert-northmind (commented):
Nice work. I think in general looks good. I left some comments

lib/src/api/common/attribute.dart
harshitt13: ahh my bad, that is actually a bad comment

lib/src/api/common/attribute_create.dart
harshitt13: you're right, and I've reversed this completely

lib/src/api/logs/logger.dart
harshitt13: agreed

lib/src/api/logs/logger.dart
harshitt13: good catch

lib/src/api/common/any_value.dart
harshitt13: You're right, Other Values says anything without a dedicated mapping SHOULD go to `string_value` via `toString()`, then to an empty `AnyValue` if that isn't possible. I was arguing from a general dislike of silent stringification rather than from the text.

Your suggestion also fixes what actually bothered me about it. Reporting through `OTelErrorHandling` means we keep the value *and* surface it, instead of dropping it.

Done:
* Unknown types now stringify via `toString()`.
* A throwing `toString()` is caught, reported, and becomes an empty `AnyValue`, which is the last resort that section prescribes. Tests cover both paths, including a fixture whose `toString()` throws.
* I skipped the intermediate `bytes_value` step, since Dart has no general way to serialise an arbitrary object to bytes.
* Still refused: a non-String map key, and a structure deeper than 32 levels. Both throw, and `attrsFromMap` drops and reports as before.

CHANGELOG updated, since removing the `.toString()` fallback was listed there as a breaking change and no longer is. I'll also rewrite the "Strict vs permissive conversion" note in the PR description, which now says the opposite of what the code does.

One thing worth knowing: a class with no custom `toString()` stringifies to `Instance of 'MyClass'`, which is rarely useful telemetry. That's the spec-prescribed outcome, so I've documented it on `fromObject` rather than trying to detect it. If you'd like a report on that specific case, it's easy to add.

Separately, I noticed the same doc says non-String map keys MUST be converted via `toString()`, which we still refuse. That's a different rule with its own test surface, so I've left it for now. Happy to do it here or as a follow-up, whichever you prefer.

lib/src/api/common/any_value.dart
```diff
@@ -0,0 +1,399 @@
...
+      // microseconds whenever the DateTime happens to carry them.
+      return AnyValueString(Timestamp.dateTimeToString(obj));
+    } else {
+      // "Mapping Arbitrary Data to OTLP AnyValue", Other Values: anything not
+      // listed above SHOULD be converted to a string via toString(), and to an
+      // empty AnyValue if that is not possible. Dart offers no general way to
+      // serialize an arbitrary object to bytes, so the spec's intermediate
+      // bytes_value step has nothing to implement and is skipped.
+      try {
+        return AnyValueString(obj.toString());
```
robert-northmind, line 206: Do we maybe have a small bug here now?
We preserve an existing `AnyValue` in the logger helper, but `_fromObject` still stringifies it. 
For example:

```dart
final original = OTelAPI.attributeInt('old', 42);
final attrs = Attributes.of({'renamed': original.value});
```

This stores the string `'42'`, so `getInt('renamed')` returns `null`. And a wrapped `AnyValueBytes([1, 2])` inside a map becomes `'<2 bytes>'`, losing the actual bytes.
I think we need the same `AnyValue` check in the recursive converter too, returning it unchanged before the fallback. 
Could we also cover attribute reinsertion and wrapped values inside maps/lists in the tests?

robert-northmind (commented):
Hey, I found one more issue.
But after that I think this looks good.
And maybe we should also let @michaelbushe have one look at this before merging. It is a bit bigger changes, and he did mention he also wanted to look it over.

robert-northmind (approved):
Nice work! Looks good for me now.
But maybe also give @michaelbushe  a chance to review before merging

PR #124, opened 2026-09-02
fix: natively honor the isRecording flag in APISpan and evaluate APITracer.isEnabled() (#96)

### LINKED ISSUE #96: [BUG] Trace API: APISpan ignores the isRecording creation parameter
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#120** — the SDK's Span wrapper enforces the recording gate today; the API type should honor it natively.

`APISpan` ignores the `isRecording` value provided at creation: mutating operations (setAttribute, addEvent, addLink, setStatus, updateName) record data regardless. Per the Trace API/SDK specs, a non-recording span MUST NOT record telemetry — mutations on it are no-ops and `isRecording` reflects the creation decision.

The SDK currently wraps `APISpan` and gates all mutators on its own recording state (see the SDK sampling work). Once the API type honors `isRecording` itself, the SDK wrapper's gate becomes a redundant safety net and API-direct users (no SDK installed, or custom SDKs) get spec behavior for free.

harshitt13:
Fixes #96 

This PR addresses the issue where `APISpan` ignored the `isRecording` value provided at creation, and also resolves a latent bug in `APITracer` that this fix uncovered. 

### Changes

- **`APISpan` natively honors `isRecording`**: Modified `APISpan` and `APISpanCreate` so that the `isRecording` state is preserved and correctly checked. If a span is created as non-recording, all mutating operations (`setAttribute`, `addEvent`, etc.) are now correctly ignored as no-ops per the Trace API specs. `NonRecordingSpan` explicitly initializes with `isRecording: false`.
- **`APITracer.isEnabled()` correctly checks API factory status**: During the fix for #96, it was discovered that `APITracer.isEnabled()` was hardcoded to unconditionally return `false`. With `isRecording` wired up, this meant any span created via the API Tracer directly became non-recording. This PR updates `isEnabled()` to correctly return `!OTelFactory.otelFactory!.isAPIFactory`, ensuring that direct use of the API Tracer behaves properly without inadvertently marking all spans as non-recording.
- **Test coverage**: Added new unit tests confirming non-recording span behavior, updated `tracer_test.dart` to expect the tracer to correctly reflect its enabled state under `SdkLikeFactory`, and maintained 100% test coverage.


lib/src/api/trace/tracer.dart
```diff
@@ -72,7 +72,12 @@ class APITracer {
   /// No parameters are currently required by the spec, but this is a method
   /// (not a getter) so parameters such as [kind] and [context] can be added
   /// later without a breaking change.
-  bool isEnabled({SpanKind? kind, Context? context}) => false;
+  bool isEnabled({SpanKind? kind, Context? context}) {
+    // No factory means nothing can record, and error-handling.md forbids
+    // throwing from an API method, so never dereference a null factory here.
+    final factory = OTelFactory.otelFactory;
```
michaelbushe, line 78: This is the part that decided it in your favour over #138, so it is worth spelling out.

Both PRs found the same coupling: once `APISpan` honours the creation flag, `createSpan`'s `(isRecording ?? true) && isEnabled(...)` starts mattering, and a base `isEnabled` hardcoded to `false` turns every span non-recording. #138 fixed it by deleting the `&& isEnabled(...)` conjunct. You fixed the hardcoded `false` and kept the conjunct.

Yours is the correct one. `trace/sdk.md` says `Enabled` MUST return `false` when there are no registered span processors, so it is a correctness gate rather than a performance hint. A tracer that MUST report disabled while still handing out recording spans is the contradiction the conjunct is there to prevent.

I did change the body. It was `!OTelFactory.otelFactory!.isAPIFactory`, and that bang throws once `reset()` has cleared the factory, which `OTelAPI.reset()` and `OTelFactory.reset()` both do. `error-handling.md` says an API method MUST NOT throw when used incorrectly, so it now returns `false` for a null factory. There is a test on it, and I checked it fails if the bang goes back.

test/unit/api/trace/tracer_test.dart
```diff
@@ -28,13 +28,27 @@ void main() {
...
       final tracer = OTelAPI.tracer('test-tracer');
 
       expect(tracer.name, equals('test-tracer'));
       expect(tracer.version, isNull);
       expect(tracer.schemaUrl, isNull);
-      expect(tracer.isEnabled(), isFalse);
+      // The test harness installs SdkLikeFactory, which sets isAPIFactory = false.
+      // APITracer natively checks !isAPIFactory to determine if it is enabled.
+      // Therefore, in this test environment, the tracer is correctly enabled.
+      expect(tracer.isEnabled(), isTrue);
```
michaelbushe, line 51: Flipping an existing assertion is the kind of thing I want to look at twice, so for the record: this one is right. The harness installs `SdkLikeFactory`, whose `isAPIFactory` is `false`, so a tracer really is enabled here and the old `isFalse` was asserting the bug. Good call explaining it in a comment rather than just changing the expectation.

michaelbushe (approved):
Nicely done, thank you.

PR #125, opened 2026-09-02
fix: remove warnings and validation from no-op MeterProvider

### LINKED ISSUE #71: [BUG] Metrics API: no-op MeterProvider logs warnings and validates arguments
> **Migrated from MindfulSoftwareLLC/dartastic_opentelemetry#146** — the fix belongs entirely in `dartastic_opentelemetry_api`; the SDK issue is closed as moved.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Metrics No-Op, MeterProvider
**URL:** https://opentelemetry.io/docs/specs/otel/metrics/noop/#meterprovider
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/noop.md#meterprovider

> The No-Op is used by OpenTelemetry users to disable OpenTelemetry computation overhead and eliminate OpenTelemetry related output. For this reason, the MeterProvider MUST NOT
> return a non-empty error or log any message for any operations it performs.

The Meter Creation section adds:

> the MeterProvider MUST NOT validate any argument it receives.

## The problem

`getMeter` on the no-op meter provider writes warnings in two places.

The empty-name check also breaks the second rule. A no-op must not validate an argument.

A user selects the no-op to remove all OpenTelemetry output. These log lines defeat that purpose.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/metrics/meter_provider.dart:81-90`

```dart
if (_isShutdown) {
  OTelLog.warn('getMeter called after shutdown; returning a no-op meter.');
}

// Validate the meter name; if invalid (empty), log a warning and use empty string.
final validatedName = name.isEmpty ? '' : name;
if (validatedName.isEmpty) {
  OTelLog.warn(
      'Invalid meter name provided; using empty string as fallback.');
}
```

## How to correct it

1. Remove both `OTelLog.warn` calls.
2. Remove the name check.
3. Accept every argument without a message, and return a no-op meter.
4. Move the shutdown warning and the name fallback log to the SDK meter provider.

## Notes

- Source finding files: `fable-ultra/metrics-api-04-noop-provider-logs.md`
- Parent issue: #95

harshitt13:
Resolves #71
This PR updates the `APIMeterProvider` to comply strictly with the OpenTelemetry Metrics No-Op specification. According to the spec (v1.60.0):
- A `MeterProvider` MUST NOT return a non-empty error or log any message for any operations it performs.
- A `MeterProvider` MUST NOT validate any argument it receives.
## Changes Made
- **Removed shutdown warning:** Removed the `OTelErrorHandling.report` call that was triggered when `getMeter` was called after shutdown.
- **Removed name validation:** Removed the logic that validated against empty meter names and logged an `ArgumentError` warning. The provider now safely accepts any argument and returns a no-op meter.
- **Updated Tests:** Renamed and updated the test `creates meter with empty name` in `meter_provider_test.dart` to assert that no fallback behavior triggers and empty names are correctly processed.
- **Added Changelog Entry:** Added a bug fix entry under `1.0.0-rc.4-wip`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.4-wip]
 
+### Fixed
+- `APIMeterProvider.getMeter` on the no-op API meter provider no longer validates arguments or logs warnings (e.g., when called after shutdown or with an empty name), complying with the OpenTelemetry Metrics No-Op spec requirements
+([#71](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/issues/71)).
+
 ## [1.0.0-rc.3] - 2026-08-27
 
 ### Changed
--- a/test/unit/api/metrics/meter_provider_test.dart
+++ b/test/unit/api/metrics/meter_provider_test.dart
@@ -30,7 +30,7 @@ void main() {
       expect(meter.schemaUrl, isNull);
     });
 
-    test('creates meter with empty name as fallback if name is invalid', () {
+    test('creates meter with empty name', () {
       // Arrange & Act
       final meter = meterProvider.getMeter(name: '');
 
--- a/lib/src/api/metrics/meter_provider.dart
+++ b/lib/src/api/metrics/meter_provider.dart
@@ -4,7 +4,6 @@
 // ignore_for_file: unnecessary_getters_setters
 
 import 'package:meta/meta.dart';
-import '../../util/otel_error_handler.dart';
 import '../common/attributes.dart';
 import '../common/signal_instance_key.dart';
 import '../otel_api.dart';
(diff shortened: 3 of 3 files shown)


yuzurihaaa (approved):

test/unit/api/metrics/meter_provider_test.dart
```diff
@@ -30,7 +30,7 @@ void main() {
       expect(meter.schemaUrl, isNull);
     });
 
-    test('creates meter with empty name as fallback if name is invalid', () {
+    test('creates meter with empty name', () {
```
michaelbushe: This asserts the meter is created but not that nothing was reported, which is the behavior the PR changes. Put the two `OTelErrorHandling.report` calls back and the suite still passes.

Install a handler and assert it stays empty, once for the empty name and once for `getMeter` after `shutdown()`, since those are the two paths you removed:

```dart
final reported = <Object>[];
OTelAPI.setErrorHandler((e, _) => reported.add(e));
addTearDown(() => OTelAPI.setErrorHandler(null));
```

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.4-wip]
 
+### Fixed
+- `APIMeterProvider.getMeter` on the no-op API meter provider no longer validates arguments or logs warnings (e.g., when called after shutdown or with an empty name), complying with the OpenTelemetry Metrics No-Op spec requirements
+([#71](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/issues/71)).
```
michaelbushe: Link the PR, `([#125](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/125))`, not the issue. Same note as on #113, #123 and #130; the entry points at the PR and the PR points at the issue.

michaelbushe:
Closing as this was folded into #113.  Thanks @harshitt13 and sorry for the confusion I may have caused.

Issue #66, opened 2026-08-23
[BUG] Trace API: APISpan exposes attributes, events, links, and status


michaelbushe:
> **Migrated from MindfulSoftwareLLC/dartastic_opentelemetry#115** — the fix belongs entirely in `dartastic_opentelemetry_api`; the SDK issue is closed as moved.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Trace API, Span
**Requirement level:** SHOULD NOT
**URL:** https://opentelemetry.io/docs/specs/otel/trace/api/#span
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#span

> `Span`s are not meant to be used to propagate information within a process. To
> prevent misuse, implementations SHOULD NOT provide access to a `Span`'s
> attributes besides its `SpanContext`.

## The problem

The API span gives a user read access to the recorded data of the span.

The `attributes` getter carries `@visibleForTesting`. That annotation is a hint for the analyzer only. The getter stays in the public API. Any code can call it. A dynamic call produces no warning at all.

Four more getters expose recorded data with no guard.

This design invites the misuse that the spec describes. Instrumentation can read data back off a live span to move information inside a process.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/trace/span.dart:158`

```dart
@visibleForTesting
Attributes get attributes => _attributes;
```

- `span.dart:144` — `List<SpanEvent>? get spanEvents`
- `span.dart:148` — `List<SpanLink>? get spanLinks`
- `span.dart:120` — `SpanStatusCode get status`
- `span.dart:123` — `String? get statusDescription`

## How to correct it

Remove these getters from `APISpan`. Move `attributes`, `spanEvents`, `spanLinks`, and the status accessors to an SDK-internal readable type.

The SDK can expose a `ReadableSpan` or a `ReadWriteSpan` for processors and exporters. The SDK spec expects read access at that level. The public API span then exposes `spanContext`, `isRecording`, and the mutating operations only.

CAUTION: This change removes public getters. It breaks callers. Add the `breaking-change` label if triage confirms the violation.

## Notes

- Spec version: **v1.60.0**. The audit ran against v1.59.0. The quoted text above was confirmed word for word in the v1.60.0 source on 2026-08-15.
- Found by: Fable 5 (x-high run)
[...]

PR #127, opened 2026-09-03
fix(baggage): getAllValues() returns name/value pairs per spec (#78)

### LINKED ISSUE #78: [BUG] Baggage API: getAllValues does not return name and value pairs
> **Migrated from MindfulSoftwareLLC/dartastic_opentelemetry#201** — the fix belongs entirely in `dartastic_opentelemetry_api`; the SDK issue is closed as moved.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Baggage API, Get All Values
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/baggage/api/#get-all-values
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/baggage/api.md#get-all-values

> Returns the name/value pairs in the `Baggage`. The order of name/value pairs MUST NOT be significant.

## The problem

`getAllValues()` returns a list of values. The spec operation returns name and value pairs.

A caller therefore cannot learn which name belongs to which value from this method.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/baggage/baggage.dart:43-45`

## How to correct it

Two options work.

One option documents `getAllEntries()` as the Get All Values operation of the spec, and deprecates the value-only list.

The other option changes the return type of `getAllValues()` to entries.

CAUTION: Changing the return type breaks callers. Add the `breaking-change` label if triage confirms the violation and the team picks that option.

## Notes

- Found by: Grok 4.5
- Source finding files: `grok/baggage/001-get-all-values-shape.md`
- Parent issue: #95

harshitt13:
Fixes #78

Spec: [Baggage API, Get All Values](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/baggage/api.md#get-all-values) (v1.60.0, requirement level MUST)

## Design choice: breaking change over deprecation

The issue outlined two valid options: deprecate `getAllValues()` in favor of `getAllEntries()`, or fix `getAllValues()`'s return type directly. I went with the breaking change (changing `getAllValues()`'s return type from `List<String>` to `Map<String, BaggageEntry>`) rather than deprecating it, for one reason: this package is still `1.0.0-rc.4-wip`, pre-stable. Breaking getAllValues()` now costs whatever few callers exist today; deprecating it instead leaves an incorrectly-shaped method in the public API through a full release cycle before it can be removed. Given #66 already introduces a confirmed breaking change in this same release, consolidating breakage into fewer version bumps seemed preferable to trickling it out. Happy to switch to the deprecation approach if you'd rather play it safer here.

**Labeled breaking-change** per the issue's own guidance.

## Changes

- `lib/src/api/baggage/baggage.dart`: `getAllValues()` now returns `Map<String, BaggageEntry>` (the same data `getAllEntries()` already returned correctly), with its doc comment updated to state it implements the spec's "Get All Values" operation.
- `test/unit/api/baggage/baggage_test.dart`: added a dedicated `getAllEntries()` test, updated the existing `getAllValues()` test's  assertions for the new return type.
- `CHANGELOG.md`: `### Changed` entry under `1.0.0-rc.4-wip` with migration guidance, referencing #78.

## Note

`getAllValues()` and `getAllEntries()` are now functionally identical (same signature, same data). That's an inherent consequence of fixing `getAllValues()` in place rather than removing it — happy to make one delegate to the other for DRYness if you'd like, just didn't want to make that call unilaterally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.4-wip]
 
+### Changed
+
+- **BREAKING**: `Baggage.getAllValues()` now returns a `Map<String, BaggageEntry>` instead of a `List<String>`. This correctly implements the OpenTelemetry Baggage API's "Get All Values" operation which requires returning the name/value pairs. Existing callers should migrate by expecting a map instead of a list.
+  ([#78](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/issues/78))
+
 ## [1.0.0-rc.3] - 2026-08-27
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- **BREAKING**: `Baggage.getAllValues()` now returns a `Map<String, BaggageEntry>` instead of a `List<String>`. This correctly implements the OpenTelemetry Baggage API's "Get All Values" operation which requires returning the name/value pairs. Existing callers should migrate by expecting a map instead of a list.
-  ([#78](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/issues/78))
+- **BREAKING**: `Baggage.getAllValues()` now returns
+  `Map<String, BaggageEntry>` instead of `List<String>`. Callers that
+  expected a list of values should read the map instead.
+  ([#127](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/127),
+  fixes [#78](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/issues/78))
+
 ### Fixed (spec compliance)
 
 - The logs API now documents that `APILoggerProvider` and `APILogger`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - **BREAKING**: `Baggage.getAllValues()` now returns
-  `Map<String, BaggageEntry>` instead of `List<String>`. Callers that
-  expected a list of values should read the map instead.
-  ([#127](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/127),
-  fixes [#78](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/issues/78))
+  `Map<String, BaggageEntry>` instead of `List<String>`, the name/value pairs
+  the specification's "Get All Values" operation requires. To get the old
+  `List<String>` back, write
+  `getAllValues().values.map((e) => e.value).toList()`.
+  `Baggage.getAllEntries()` is now `@Deprecated` and delegates to
+  `getAllValues()`; it will be removed in a future release
+  ([#127](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/127)).
 
 ### Fixed (spec compliance)
 
--- a/lib/src/api/baggage/baggage.dart
+++ b/lib/src/api/baggage/baggage.dart
@@ -36,14 +36,19 @@ class Baggage {
   BaggageEntry? getEntry(String key) => _entries[key];
 
   /// Returns all entries in this Baggage as an immutable map.
+  ///
+  /// This implements the "Get All Values" operation from the OpenTelemetry
+  /// Baggage specification, which requires returning the name/value pairs.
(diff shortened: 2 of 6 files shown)


lib/src/api/baggage/baggage.dart
```diff
@@ -36,14 +36,19 @@ class Baggage {
...
   /// Retrieves a Baggage value for the given key, or `null` if not present.
   String? getValue(String key) => _entries[key]?.value;
 
-  /// Retrieves all Baggage values for the given key, or `null` if not present.
-  List<String> getAllValues() =>
-      _entries.values.map((entry) => entry.value).toList();
+  /// Returns all name/value pairs in this Baggage as an immutable map.
+  ///
+  /// This implements the "Get All Values" operation from the OpenTelemetry
+  /// Baggage specification, which requires returning the name/value pairs.
```
robert-northmind: Does this part of the documentation bring any real content? We sort of know already that we are following the OTel Sec in this repo.
Maybe we could just drop this part and only keep the
`Returns all name/value pairs in this Baggage as an immutable map.`?

The same comment to the other doc change above.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.4-wip]
 
+### Changed
+
+- **BREAKING**: `Baggage.getAllValues()` now returns a `Map<String, BaggageEntry>` instead of a `List<String>`. This correctly implements the OpenTelemetry Baggage API's "Get All Values" operation which requires returning the name/value pairs. Existing callers should migrate by expecting a map instead of a list.
```
robert-northmind: Should we minimize this changelog entry a bit? I think some patterns we are going for in the changelog is to keep it clean and to the core point.
Maybe the fact that we correctly implement the OTel Baggage API is not needed here. You would also find that detail by clicking the issue-link.

What do you think?

robert-northmind (commented):
Nice! Looks great. Just left a few minor comments on some code comments

robert-northmind (approved):

lib/src/api/baggage/baggage.dart
```diff
@@ -41,9 +41,8 @@ class Baggage {
   /// Retrieves a Baggage value for the given key, or `null` if not present.
   String? getValue(String key) => _entries[key]?.value;
 
-  /// Retrieves all Baggage values for the given key, or `null` if not present.
-  List<String> getAllValues() =>
-      _entries.values.map((entry) => entry.value).toList();
+  /// Returns all name/value pairs in this Baggage as an immutable map.
+  Map<String, BaggageEntry> getAllValues() => _entries;
```
michaelbushe, line 48: This is the same as `getAllEntries()` six lines up. Same signature, same body, same meaning.

#78 offered two routes, deprecate the value-only list in favour of `getAllEntries()`, or change the return type. This does the second but leaves the first one's target sitting there, so we end up with both.

Pick one and deprecate the other. I would keep `getAllValues()`, since that is the name the spec operation uses, and mark `getAllEntries()` deprecated pointing here. That way existing `getAllEntries()` callers get a migration note instead of silently having two identical options forever.

test/unit/api/baggage/baggage_test.dart
```diff
@@ -58,8 +73,9 @@ void main() {
       });
 
       final values = baggage.getAllValues();
-      expect(values, containsAll(['value1', 'value2']));
       expect(values.length, equals(2));
+      expect(values['key1'], equals(entry1));
+      expect(values['key2'], equals(entry2));
     });
```
michaelbushe: Similarly, these two tests are now the same test written twice, differing only in which of the two identical methods they call. That is the duplication showing up in the suite.

Once one of the methods is deprecated, this collapses to one test on the surviving method, plus a small one asserting the deprecated name still delegates to it.

CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [1.0.0-rc.4-wip]
 
+### Changed
+
+- **BREAKING**: `Baggage.getAllValues()` now returns
+  `Map<String, BaggageEntry>` instead of `List<String>`. Callers that
+  expected a list of values should read the map instead.
+  ([#127](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/127),
+  fixes [#78](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/issues/78))
```
michaelbushe: "should read the map instead" does not tell anyone what to write. Give them the line: `getAllValues().values.map((e) => e.value)` gets the old `List<String>` back.

Also drop the issue link and keep the PR one. The entries around this all carry the PR only.

michaelbushe (changes requested):

michaelbushe (approved):
Looks good now, thank you.

PR #128, opened 2026-09-03
Apply W3C tracestate truncation in toHeaderString

### LINKED ISSUE #126: [BUG] Trace API: tracestate truncation ignores the 512-character budget and the 128-character rule
## Spec requirement

**Spec:** W3C Trace Context, reached through OpenTelemetry Specification v1.60.0
**Requirement level:** MUST (truncate whole entries) and SHOULD (the rest)
**Pinned source:** https://www.w3.org/TR/trace-context/#tracestate-limits

`TraceState` is an API type (`specification/trace/api.md:278`), and the spec delegates its rules rather than restating them:

> These operations MUST follow the rules described in the [W3C Trace Context specification](https://www.w3.org/TR/trace-context/#mutating-the-tracestate-field).

`specification/trace/api.md:289`.

W3C §3.3.1.5 then says:

> Vendors _SHOULD_ propagate at least 512 characters of a combined header.

> the vendor _MUST_ truncate whole entries. Entries larger than `128` characters long _SHOULD_ be removed first. Then entries _SHOULD_ be removed starting from the end of `tracestate`.

## The problem

We implement one third of the truncation procedure, and we drive it off the wrong trigger.

| Rule | Level | Implemented |
|---|---|---|
| Truncate whole entries | MUST | yes |
| Entries over 128 characters removed **first** | SHOULD | **no** |
| Then remove starting from the end | SHOULD | yes |
| Propagate at least 512 characters of a combined header | SHOULD | **no** |

There is no `512` and no 128-character handling anywhere in `lib/`. What we do instead is cap the member count at 32 (W3C §3.3.1.1, "There can be a maximum of 32 `list-member`s in a `list`") and drop from the end when a `put` would exceed it.

The 32 cap is correct on its own terms, but it is a grammar limit on the list, not the truncation procedure. A `tracestate` with 32 short members can still exceed 512 characters and go out oversized. And when something does have to go, we drop a small oldest entry while keeping a 300-character one that the spec says should have gone first.

"Combined header" here means the `tracestate` header value itself, including the case where it arrives split across repeated header instances and is recombined. It is not traceparent plus tracestate. So this is measurable from `TraceState` alone and does not need any propagator context.

## Offending code

The member cap is enforced in two places and the character budget in neither:

- `lib/src/api/trace/trace_state.dart:43` (`fromString`)
- `lib/src/api/trace/trace_state.dart:95` (`put`)

`toString()` at `lib/src/api/trace/trace_state.dart` serializes with no length check:

```dart
String toString() {
  return _entries.entries.map((e) => '${e.key}=${e.value}').join(',');
}
```

## How to correct it

This belongs here rather than in a propagator. The 32-member cap already lives in this class, `toString()` is the serializer, and every propagator that emits a `tracestate` header should get the same behavior without reimplementing it. See dartastic_opentelemetry#295 for the SDK half, where the W3C propagator currently duplicates this serializer instead of calling it.

1. Apply §3.3.1.5 when producing the header value: remove whole entries only, entries over 128 characters first, then from the end.
2. Keep the existing 32-member cap. It is a separate rule and it is correct.
3. Decide whether `toString()` truncates or a dedicated method does. A getter that silently drops data is surprising, so a separate method the propagator calls may be the better shape. Worth settling in this issue before implementation.
4. Report dropped entries through `OTelErrorHandling`, matching how invalid entries are handled today.

Open question worth resolving here: 512 is a floor on what vendors _should propagate_, not a ceiling we are obliged to impose. Truncating to 512 is the conservative reading. Whether the budget should be configurable is undecided.

## Notes

- Found while reviewing #115, which corrected the mutation ordering so that "remove from the end" now genuinely removes the oldest. That made the missing half of the procedure visible.
- The W3C text above was read from https://www.w3.org/TR/trace-context/ on 2026-09-03.
- Related: #61 (put ordering), #62 (construction validation), #93 (parse validation).

i-am-paradox:
Implements the truncation procedure from W3C Trace Context 3.3.1.5, following the shape proposed in my comment on the issue.

`TraceState.toHeaderString()` produces the header value the way the spec asks: entries over 128 characters are removed first, then entries are removed from the end until the joined value fits the 512-character budget. Only whole entries are dropped, and each dropped entry is reported through `OTelErrorHandling` the same way invalid entries already are.

`toString()` is unchanged, so it keeps serializing everything the state still holds; the unbounded and truncated forms are now distinct, and a printed state can not silently disagree with `entries`. If maintainers would rather have the budget configurable or `toString()` itself truncate, I'm open to reshaping this.

The existing 32-member cap stays where it is, in `fromString` and `put`: it is the separate 3.3.1.1 grammar limit.

Tests cover the 128-first ordering, the end-removal ordering against the 512 boundary, whole-entry removal, the `toString`/`toHeaderString` split, and the error reports. Full suite passes locally, `dart analyze` clean.

With this in place, dartastic_opentelemetry#295 can proceed with the propagator delegating to this method instead of its private copy.

Fixes #126

### DIFF
--- a/lib/src/api/trace/trace_state.dart
+++ b/lib/src/api/trace/trace_state.dart
@@ -9,6 +9,15 @@ part 'trace_state_create.dart';
 
 /// Key-value pairs carried along with a span context.
 /// TraceState follows the W3C Trace Context specification.
+///
+/// Size policy: the grammar limits (W3C §3.3.1.1) — a maximum of 32
+/// list-members and the per-key/per-value length rules — are enforced on
+/// every path. [toString] serializes exactly what the state holds and
+/// does not truncate beyond them; [toHeaderString] applies the §3.3.1.5
+/// truncation procedure for callers that need a bounded header value.
+/// Vendors SHOULD propagate at least 512 characters of the combined
+/// header, so 512 is a floor the procedure keeps whole entries within,
+/// not a ceiling imposed on the state itself.
 class TraceState {
   static const int _maxKeyValuePairs = 32;
   static final RegExp _simpleKeyFormat = RegExp(r'^[a-z][a-z0-9_\-*/]{0,255}$');
--- a/lib/src/api/trace/trace_state.dart
+++ b/lib/src/api/trace/trace_state.dart
@@ -119,6 +119,41 @@ class TraceState {
     return _entries.entries.map((e) => '${e.key}=${e.value}').join(',');
   }
 
+  /// Produces the W3C `tracestate` header value, applying the truncation
+  /// procedure of W3C Trace Context §3.3.1.5.
+  ///
+  /// Entries over 128 characters are removed first, then entries are removed
+  /// from the end until the joined value fits the 512-character budget.
+  /// Only whole entries are removed, and every dropped entry is reported
+  /// through [OTelErrorHandling]. Unlike [toString], this may return a value
+  /// that no longer contains all entries.
+  String toHeaderString() {
+    final entries = List<MapEntry<String, String>>.from(_entries.entries);
+
+    // W3C §3.3.1.5: entries larger than 128 characters should be removed
+    // first. The length of a list-member is its `key=value` size.
+    final overlong = entries
+        .where((e) => '${e.key}=${e.value}'.length > 128)
+        .toList(growable: false);
+    for (final entry in overlong) {
+      entries.remove(entry);
+      OTelErrorHandling.report(StateError(
+          'TraceState entry ${entry.key} exceeds 128 characters; dropped.'));
+    }
+
+    // Then entries should be removed starting from the end of the
+    // tracestate until the value fits the 512-character budget.
+    var value = entries.map((e) => '${e.key}=${e.value}').join(',');
+    while (value.length > 512 && entries.isNotEmpty) {
+      final entry = entries.removeLast();
+      OTelErrorHandling.report(StateError(
+          'TraceState exceeds 512 characters; entry ${entry.key} dropped.'));
+      value = entries.map((e) => '${e.key}=${e.value}').join(',');
+    }
+
+    return value;
+  }
+
   /// Validate a tracestate key: a simple key, or a multi-tenant
   /// `tenant-id@system-id` key.
   static bool _isValidKey(String key) {
--- a/lib/src/api/trace/trace_state.dart
+++ b/lib/src/api/trace/trace_state.dart
@@ -131,16 +131,23 @@ class TraceState {
   /// Produces the W3C `tracestate` header value, applying the truncation
   /// procedure of W3C Trace Context §3.3.1.5.
   ///
-  /// Entries over 128 characters are removed first, then entries are removed
-  /// from the end until the joined value fits the 512-character budget.
-  /// Only whole entries are removed, and every dropped entry is reported
-  /// through [OTelErrorHandling]. Unlike [toString], this may return a value
-  /// that no longer contains all entries.
+  /// The procedure only runs when the value needs to be truncated: if the
+  /// joined value fits the 512-character budget it is returned as-is,
+  /// including entries over 128 characters. When it does not fit, whole
+  /// entries are removed, entries larger than 128 characters first, then
+  /// entries from the end until the value fits. Every dropped entry is
+  /// reported through [OTelErrorHandling]. Unlike [toString], this may
+  /// return a value that no longer contains all entries.
   String toHeaderString() {
-    final entries = List<MapEntry<String, String>>.from(_entries.entries);
+    var value = _entries.entries.map((e) => '${e.key}=${e.value}').join(',');
+    if (value.length <= 512) {
+      return value;
+    }
 
-    // W3C §3.3.1.5: entries larger than 128 characters should be removed
-    // first. The length of a list-member is its `key=value` size.
+    // W3C §3.3.1.5: "Entries larger than 128 characters long SHOULD be
+    // removed first", as part of truncating a value that does not fit.
+    // The length of a list-member is its `key=value` size.
+    final entries = List<MapEntry<String, String>>.from(_entries.entries);
     final overlong = entries
         .where((e) => '${e.key}=${e.value}'.length > 128)
         .toList(growable: false);
@@ -148,11 +155,11 @@ class TraceState {
       entries.remove(entry);
       OTelErrorHandling.report(StateError(
           'TraceState entry ${entry.key} exceeds 128 characters; dropped.'));
+      value = entries.map((e) => '${e.key}=${e.value}').join(',');
     }
 
     // Then entries should be removed starting from the end of the
     // tracestate until the value fits the 512-character budget.
-    var value = entries.map((e) => '${e.key}=${e.value}').join(',');
     while (value.length > 512 && entries.isNotEmpty) {
       final entry = entries.removeLast();
       OTelErrorHandling.report(StateError(
--- a/test/unit/api/trace/trace_state_spec_compliance_test.dart
+++ b/test/unit/api/trace/trace_state_spec_compliance_test.dart
@@ -62,4 +62,86 @@ void main() {
       expect(result.get('vendor'), isNull);
     });
   });
+
+  group('TraceState toHeaderString follows W3C 3.3.1.5 truncation', () {
+    test('returns the full value when it fits the budget', () {
+      final traceState = TraceState.fromMap({'a': '1', 'b': '2'});
+      expect(traceState.toHeaderString(), equals('a=1,b=2'));
+    });
+
+    test('returns empty string for an empty state', () {
+      final traceState = TraceState.empty();
+      expect(traceState.toHeaderString(), equals(''));
+    });
+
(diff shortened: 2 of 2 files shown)


michaelbushe:
Merging this after one change. I'd started drafting a decline, having re-read §3.3.1.5 and concluded that every truncation rule is gated on "needs to be truncated," so an SDK isn't obliged to truncate at all. But you built it the right way: toString() unchanged means this is opt-in at the call site, and that's exactly what's needed. The default propagator will keep calling toString() and drop nothing. A Dart server forwarding through a proxy with header limits is the caller that needs toHeaderString(), and now it has it.
 
One addition before I merge, since the one SHOULD in that section that isn't gated is "the maximum size of the propagated tracestate header SHOULD be documented." Please sdd a paragraph to the TraceState class doc stating the policy: the 32-entry and 256-character grammar limits are enforced on every path, toString() does not truncate beyond them, and toHeaderString() applies the §3.3.1.5 procedure for callers that need a bounded header. Cite the "at least 512" sentence so nobody reads 512 as a ceiling later.

Thanks for the fast turnaround and for working the open questions in the issue before building. Nice first PR!

i-am-paradox:
Added the size-policy paragraph to the TraceState class doc in a8a1dc8: grammar limits on every path, the two serializers, and the 512 floor cited so it can't be read as a ceiling.

robert-northmind:
Hi, thanks for tackling this topic :) And it looks good. I just have one question:

Should we first check if the full header actually exceeds 512 before dropping any entries here?
I read [§3.3.1.5](https://www.w3.org/TR/trace-context/#tracestate-limits) as making the removal rules conditional on needing to truncate. So "big=<130 characters>" would still fit comfortably at 134 total (less than 512). but currently toHeaderString() drops it completely. It assumes we should always drop entries bigger than 128.

The spec speaks of "size limitations" and then mentions the 512.. 
and if we are in a situation where tracestate needs to be truncated due to this size limitation, then we should start removing entries larger than 128.

So should we keep the full header when it fits, then remove entries over 128 one at a time while we’re over budget, followed by entries from the end if needed? We could add a test for that 134-character case, and another where removing one large entry makes the rest fit. 

What do you think? Or am I reading the W3C spec wrong?

i-am-paradox:
You're reading the spec right. I re-read 3.3.1.5 and the gating sentence is explicit: 'In a situation where tracestate needs to be truncated due to size limitations, the vendor MUST truncate whole entries. Entries larger than 128 characters long SHOULD be removed first.' The removal rules are part of the truncation procedure, not a standing 128-character admission limit.

Reworked in 243ba9f: toHeaderString() now returns the value untouched when it fits 512, so your 134-character case keeps its big entry. Once it is over budget, the procedure runs in the order you described: over-128 entries go first (one pass, then re-join), then entries from the end until it fits.

Tests updated for both: the 134-character case asserts the entry survives, and a new case shows removing one over-128 entry bringing the rest under budget without further drops. One thing I hit while reworking the fixtures: fromString/fromMap only ever hold 32 entries (the 3.3.1.1 cap), so a >512 value needs most entries to be around 13 characters, worth knowing for anyone writing cases against these limits. Full suite passes locally (1182), analyze clean.

lib/src/api/trace/trace_state.dart
```diff
@@ -119,6 +128,48 @@ class TraceState {
...
+  /// entries are removed, entries larger than 128 characters first, then
+  /// entries from the end until the value fits. Every dropped entry is
+  /// reported through [OTelErrorHandling]. Unlike [toString], this may
+  /// return a value that no longer contains all entries.
+  String toHeaderString() {
+    var value = _entries.entries.map((e) => '${e.key}=${e.value}').join(',');
+    if (value.length <= 512) {
+      return value;
+    }
+
```
robert-northmind, line 146: Should we also stop removing entries here once the remaining value fits within 512?

The initial check handles the under-budget case now 👍 
But e.g. with three entries that each have a 180-character value, we start at 548 characters. Removing one brings us to 365, but the loop still removes the other two and returns an empty string.

The w3c spec doesn't explicitly say we have to check after each removal. But my reading is that once we fit the size limit, there's no need to drop more tracing data. What do you think?

Could we add a test with multiple large entries where removing just one is enough? The current cases only have one large entry, so they don't catch this. We could assert the exact entries that survive and that only one drop is reported.

robert-northmind (commented):
Looks good. Just left one more small comment. Let me know what you think.

And should we also add a changelog entry for the new toHeaderString() method?

lib/src/api/trace/trace_state.dart
```diff
@@ -9,6 +9,15 @@ part 'trace_state_create.dart';
...
 /// Key-value pairs carried along with a span context.
 /// TraceState follows the W3C Trace Context specification.
+///
+/// Size policy: the grammar limits (W3C §3.3.1.1) — a maximum of 32
+/// list-members and the per-key/per-value length rules — are enforced on
+/// every path. [toString] serializes exactly what the state holds and
+/// does not truncate beyond them; [toHeaderString] applies the §3.3.1.5
+/// truncation procedure for callers that need a bounded header value.
+/// Vendors SHOULD propagate at least 512 characters of the combined
+/// header, so 512 is a floor the procedure keeps whole entries within,
```
michaelbushe, line 19: The over-128 loop never checks the budget, so once truncation starts it removes every over-128 entry even after the value already fits. It recomputes `value` on line 18 but never reads it. W3C puts over-128 entries first among the removals that are needed, not unconditionally.

```dart
for (final entry in overlong) {
  if (value.length <= 512) break;
  ...
}
```

Two things on cost, since this can run once per outbound request. Rejoining the whole string after every removal is O(n²); keep a running length and join once at the end. And `entries.remove(entry)` is a linear scan inside the loop, and `MapEntry` has no value equality so it only works because these are the same instances. One pass that keeps what fits avoids both.

The comment on `trace_state_spec_compliance_test.dart` has the smallest case that fails.

test/unit/api/trace/trace_state_spec_compliance_test.dart
```diff
@@ -101,4 +101,124 @@ void main() {
...
+      final ok2Value = List.filled(58, 'x').join(); // ok2=... 62 chars
+      final traceState = TraceState.fromMap({
+        'ok1': ok1Value,
+        'ok2': ok2Value,
+        'big': bigValue,
+      });
+      final header = traceState.toHeaderString();
+      expect(header, equals('ok1=$ok1Value,ok2=$ok2Value'));
+      expect(header.length, lessThanOrEqualTo(512));
+    });
```
michaelbushe, line 159: None of these tests has more than one entry over 128 characters. The loop that removes over-128 entries only goes wrong when there are two or more, so it never gets exercised.

Smallest failing case:

```dart
test('a header one character over budget keeps what fits', () {
  // a=... is 255, b=... is 257, plus the comma: 513.
  final ts = TraceState.fromMap({'a': 'x' * 253, 'b': 'y' * 255});
  expect(ts.toHeaderString(), isNotEmpty);
});
```

Today this returns an empty string. Both entries are over 128, so both are removed, when removing either one alone would fit.

Other cases with no coverage yet, all passing today:

- returns the value untouched at exactly 512 characters
- an entry of exactly 128 characters is not over-long
- a single maximum-size entry yields an empty header (256 key plus 256 value is 513, nothing can fit)
- truncation leaves the TraceState itself unchanged
- surviving entries keep their relative order
- reports once per dropped entry when both removal passes run

michaelbushe (changes requested):

michaelbushe:
@robert-northmind is right, and it is worse than his example suggested. I ran his case against this branch:

```
full length          : 551
one entry length     : 183
after removing one   : 367
toHeaderString len   : 0
toHeaderString value : ""
```

Three entries, 551 characters, and the method hands back an empty string. Removing one would have brought it to 367 and fit with room to spare. Detail in the review comments.

Everything else here is good. The under-budget early return in 243ba9f is exactly right, and the class doc you added is the thing that makes this safe to have in the API at all.

One more thing, separate from this PR. The SDK propagator does not call `TraceState` for serialization at all today; it has a private copy of `toString()` in `_serializeTracestate`. That is dartastic_opentelemetry#295, and it is the reason `toHeaderString()` has no caller yet: the path that would use it bypasses the API entirely. If you want it, it is yours. It is a small change and it is the one that connects this work to something that runs.

i-am-paradox:
Fixed in 33123ef. Both removal passes now check the budget before each drop, so the five-entry case keeps the two entries that still fit, and the length is tracked arithmetically with a single join instead of a rejoin per removal.

Tests added: your two-entry 513-character case, a three-overlong case asserting exactly one report, an exactly-512 case, five 128-character entries proving the boundary, and the single max-size entry yielding an empty header. Full suite 1222 passing.

robert-northmind (commented):
Hi, nice work 🚀  thanks for all you effort.
Just 2 small things I would still consider:
1: Should we also cover some of the test cases Michael mentioned? One where both removal passes run, asserting the exact surviving entries and their order, one report per dropped entry, and that the original TraceState.entries stays unchanged.

2: Could we add a changelog entry for toHeaderString()? Since this is a new method. Maybe we should document it there

PR #129, opened 2026-09-03
fix(trace): avoid re-wrapping non-recording spans, fix scope attribute leak (#63, #65)

### LINKED ISSUE #63: [BUG] Trace API: span creation without an SDK makes a new NonRecordingSpan
> **Migrated from MindfulSoftwareLLC/dartastic_opentelemetry#110** — the fix belongs entirely in `dartastic_opentelemetry_api`; the SDK issue is closed as moved.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Trace API, Behavior of the API in the absence of an installed SDK
**Requirement level:** SHOULD
**URL:** https://opentelemetry.io/docs/specs/otel/trace/api/#behavior-of-the-api-in-the-absence-of-an-installed-sdk
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#behavior-of-the-api-in-the-absence-of-an-installed-sdk

> The API MUST return a non-recording `Span` with the `SpanContext` in the parent `Context` (whether explicitly given or implicit current).
> If the `Span` in the parent `Context` is already non-recording, it SHOULD be returned directly
> without instantiating a new `Span`.

## The problem

The parent `Context` can already hold a non-recording span. Two common cases are a `NonRecordingSpan` that wraps a `SpanContext` from a propagator, and the result of an earlier no-op `startSpan` call.

In both cases the code still creates a new `NonRecordingSpan`. Each new span also allocates an `InstrumentationScope` object and an `Attributes` object.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/trace/tracer.dart:162`

```dart
if (OTelFactory.otelFactory!.isAPIFactory) {
  final parentSpanContext = spanContext ??
      effectiveParentSpan?.spanContext ??
      contextOfSpan.spanContext;
  return NonRecordingSpan(
      parentSpanContext ?? OTelFactory.otelFactory!.spanContextInvalid());
}
```

## How to correct it

Return the parent directly when the parent is already non-recording:

```dart
if (OTelFactory.otelFactory!.isAPIFactory) {
  final parent = effectiveParentSpan;
  if (spanContext == null && parent != null && !parent.isRecording) {
    return parent; // already non-recording: return directly per spec
  }
  final parentSpanContext =
      spanContext ?? parent?.spanContext ?? contextOfSpan.spanContext;
  return NonRecordingSpan(
      parentSpanContext ?? OTelFactory.otelFactory!.spanContextInvalid());
}
```

## Notes

- Source finding files: `fable-ultra/trace-api-05-noop-wraps-new-nonrecording-span.md`, `fable-medium/trace-api-noop-wraps-nonrecording-parent.md`
- Parent issue: #95

### LINKED ISSUE #65: [BUG] Trace API: span attributes become the Tracer scope attributes
> **Migrated from MindfulSoftwareLLC/dartastic_opentelemetry#114** — the fix belongs entirely in `dartastic_opentelemetry_api`; the SDK issue is closed as moved.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Trace API, Get a Tracer
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/trace/api/#get-a-tracer
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#get-a-tracer

> This API MUST accept the following parameters: ... `attributes` (optional): Specifies the instrumentation scope attributes to associate with emitted telemetry.
>
> `version` (optional): Specifies the version of the instrumentation scope if the scope has a version (e.g. a library version).

## The problem

`createSpan` builds the instrumentation scope from its own `attributes` parameter. That parameter holds the initial attributes of the span.

As a result, two faults occur. The scope attributes from `getTracer(name, attributes: ...)` are dropped. The per-span attributes leak into the scope identity of the telemetry.

`createSpan` also invents `version: '1.0.0'` when the tracer has no version. This reports a scope version that nobody specified, and it breaks scope identity.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/trace/tracer.dart:230-246`

```dart
final apiSpan = APISpanCreate.create(
  name: name,
  instrumentationScope: InstrumentationScopeCreate.create(
      name: this.name,
      version: version ?? '1.0.0',
      schemaUrl: schemaUrl,
      attributes: attributes),   // <-- createSpan's span attributes, shadows this.attributes
  ...
```

## How to correct it

Pass `attributes: this.attributes` when the code builds the `InstrumentationScope`. Keep the span attributes on the span only. Pass `version: version` with no default, so a missing version stays null.

Build the `InstrumentationScope` one time for each tracer. Do not build it for each span.

NOTE: This issue and the `getTracer` defaults issue share a root cause. Cross-link them at triage.

## Notes

- Found by: Fable 5 (medium run)
- Source finding files: `fable-medium/trace-api-scope-attributes-from-span.md`
- Parent issue: #95

harshitt13:
Fixes #63
Fixes #65 

### #63 — span creation without an SDK re-wraps an already non-recording span

Spec: [Behavior of the API in the absence of an installed SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#behavior-of-the-api-in-the-absence-of-an-installed-sdk) (v1.60.0, SHOULD)

> If the Span in the parent Context is already non-recording, it SHOULD
> be returned directly without instantiating a new Span.

When no SDK is installed, `createSpan` always constructed a fresh `NonRecordingSpan`, even when the parent context already held one (either a `NonRecordingSpan` wrapping a propagated `SpanContext`, or the result of an earlier no-op `startSpan` call). Now it returns the existing non-recording parent directly when one is present.

### #65 — span attributes leak into the tracer's scope identity

Spec: [Get a Tracer](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#get-a-tracer) (v1.60.0, MUST)

`createSpan` built a fresh `InstrumentationScope` on every call, using the per-span `attributes` parameter instead of the tracer's own attributes from `getTracer(...)`. This dropped the real scope attributes and let span-level attributes leak into scope identity. It also defaulted `version` to `'1.0.0'` when none was supplied, fabricating a version nobody specified. `InstrumentationScope` is now built exactly once, at tracer construction, from the tracer's own `name`/`version`/`schemaUrl`/ `attributes`, with no default applied to `version`.

## Cross-repo note

The `InstrumentationScope` cache backing the #65 fix is a **private** field on `APITracer`. `dartastic_opentelemetry`'s `Tracer` class does `implements APITracer` rather than `extends`, and manually re-exposes every interface member via a delegating getter — a public field here would have required a matching SDK-side change to avoid a compile break. Kept it private since #65 doesn't require exposing it as new public API surface; nothing needed on the SDK side for this one.

## Changes

- `lib/src/api/trace/tracer.dart`:
  - `createSpan` returns the parent directly when it's already non-recording (API-only path).
  - `APITracer` now builds and caches `_instrumentationScope` once in its constructor; `createSpan` reuses it instead of rebuilding per span.
- `test/unit/api/trace/no_sdk_span_behavior_test.dart`: new test covering both non-recording-parent cases from #63, asserting   `identical()` (not just equality) on the returned span.
- `test/unit/api/trace/tracer_test.dart`: new tests asserting the scope reflects tracer attributes (not span attributes), and that the same `InstrumentationScope` instance is shared across spans from one tracer.
- `CHANGELOG.md`: entries for both issues under `1.0.0-rc.4-wip`


robert-northmind (approved):
Nice! LGTM

michaelbushe (approved):

PR #130, opened 2026-09-03
Fix: Remove Context.copyWithValue to ensure spec-compliant key retrieval (#79)

### LINKED ISSUE #79: [BUG] Context API: copyWithValue generates a key the caller can never use
> **Migrated from MindfulSoftwareLLC/dartastic_opentelemetry#202** — the fix belongs entirely in `dartastic_opentelemetry_api`; the SDK issue is closed as moved.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Context, Set value
**Requirement level:** MUST
**URL:** https://opentelemetry.io/docs/specs/otel/context/README/#set-value
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/context/README.md#set-value

> The API MUST accept the following parameters:
>
> - The `Context`.
> - The key.
> - The value to be set.
>
> The API MUST return a new `Context` containing the new value.

## The problem

`Context` carries an extra method, `copyWithValue`. It does not accept a key. It accepts a string name, then builds a new unique key inside the method.

That key is never returned to the caller. The generated key ID is therefore lost.

No caller can ever read the value back with `Context.get(key)`. The value is written into the Context and becomes unreachable.

The existing `copyWith` method is correct. It accepts a `ContextKey`.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/context/context.dart:247-255`

```dart
Context copyWithValue<T>(String name, T contextValue,
    {bool isTransferable = false}) {
  return ContextCreate.create(contextMap: {
    ..._values,
    _getAndCacheOTelFactory().contextKey<T>(
        name, ContextKey.generateContextKeyId(),
        isTransferable: isTransferable): contextValue,
  });
}
```

## How to correct it

Remove `copyWithValue` from `Context`.

A caller creates a key with `OTelAPI.contextKey<T>(name)` first. The caller then passes that key to `copyWith`, which already meets the requirement.

CAUTION: This change removes a public method. It breaks callers. Add the `breaking-change` label if triage confirms the violation.

## Notes

- Found by: Gemini 3.1 Pro (run 2)
- Source finding files: `gemini-fail-mindful/context_api_set_value_missing_key.md`
- Parent issue: #95

harshitt13:
Fixes #79

### Description

This PR addresses a spec violation where `Context.copyWithValue` dynamically generated an inaccessible key ID internally, meaning callers were never able to retrieve the stored value back via `Context.get(key)`.

To resolve this and bring the Context API fully in line with the OpenTelemetry Context spec (v1.60.0), `copyWithValue` has been removed. Callers must now explicitly construct a key using `OTelAPI.contextKey<T>(name)` and then use `Context.copyWith(key, value)`. 

### Changes
- **BREAKING:** Removed the `Context.copyWithValue` method.
- Updated all usages inside `context_coverage_test.dart` to use `OTelAPI.contextKey` and `Context.copyWith` instead.
- Retained coverage for `Context`'s internal caching mechanism (`_getAndCacheOTelFactory`) by triggering it via `Context.deserialize({})` inside `context_factory_upgrade_test.dart`.
- Added the corresponding breaking change note to `CHANGELOG.md`.


robert-northmind (approved):
LGTM 🚀

test/unit/api/context/context_factory_upgrade_test.dart
```diff
@@ -42,13 +42,12 @@ void main() {
...
+    // Context must do the same: deserialize creates its context through the
     // factory, which must now be the SDK-installed one, not the cached no-op.
-    Context.current.copyWithValue<String>('via-context', 'value');
-    expect(sdkFactory.contextKeyCalls, 2,
-        reason: 'Context must delegate to the newly installed factory, '
-            'not a no-op factory cached before the SDK initialized');
+    Context.deserialize({});
+    expect(sdkFactory.contextCalls, 1,
+        reason: 'Context.deserialize must delegate to the newly installed '
+            'factory, not a no-op factory cached before the SDK initialized');
```
michaelbushe: This drops coverage rather than moving it. The old assertion proved `Context` creates **keys** through the newly installed factory instead of a stale cached no-op, which is the bug this test exists for. Swapping to `contextCalls` proves `context()` delegates, which is a different thing, and nothing now covers the key path from inside `Context`.

`deserialize` does call `contextKey`, but only in the loop over serialized values, and `{}` never enters it. Pass one transferable entry and you get both:

```dart
Context.deserialize({
  'k': {'value': 'v', 'uniqueId': [1, 2, 3], 'originalKeyName': 'k'},
});
expect(sdkFactory.contextCalls, 1, reason: '...');
expect(sdkFactory.contextKeyCalls, 2, reason: '...');
```

That keeps the original guarantee and adds the new one.

CHANGELOG.md
```diff
@@ -14,6 +14,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   a MUST. Comments only, no behavior change
   ([#121](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/121)).
 
+### Changed
+
+- **BREAKING**: Removed `copyWithValue` from `Context` because it generated a key the caller could never retrieve. Use `OTelAPI.contextKey` and `Context.copyWith` instead
+  ([#79](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/issues/79)).
```
michaelbushe: Two things.

The link points at issue #79. Convention here is the PR, so `([#130](https://github.com/MindfulSoftwareLLC/dartastic_opentelemetry_api/pull/130))`. Standing note rather than a one-off: #113 and #123 do the same, so worth fixing in your habit rather than one at a time.

And this is a removal, so `### Removed` says it plainly. Keep a Changelog orders sections Added, Changed, Deprecated, Removed, Fixed, so it also wants to go above the `### Fixed (spec compliance)` block rather than below it.

michaelbushe (approved):
Thanks for doing this, I remember writing this originally and am shocked we had something not giving back context keys.

PR #137, opened 2026-09-11
traceIdFrom and spanIdFrom return safe defaults instead of throwing

### LINKED ISSUE #132: [BUG] Trace API: traceIdFrom and spanIdFrom throw on invalid input (API half)
> **API half.** The matching SDK issue in `dartastic_opentelemetry` will cover removing the `try/catch` that `W3CTraceContextPropagator` currently needs around `OTel.traceIdFrom` / `OTel.spanIdFrom`. That can only happen after this fix ships in an API release.
>
> Raised during review of #112. The throw itself is older than that PR.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Error handling
**Requirement level:** MUST NOT / SHOULD
**URL:** https://opentelemetry.io/docs/specs/otel/error-handling/
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/error-handling.md

>
> 1. API methods MUST NOT throw unhandled exceptions when used incorrectly by end users. The API and SDK SHOULD provide safe defaults for missing or invalid arguments.

## The problem

`OTelAPI.traceIdFrom` and `OTelAPI.spanIdFrom` throw `FormatException` when the hex string is malformed or the wrong length. #112 tightened `hexToBytes` to lowercase hex only, so the set of inputs that reach the throw grew: uppercase ids that previously decoded now throw.

Every caller has to wrap these in `try/catch` to stay spec-compliant. The SDK propagator does; `SpanContext.fromJson` does not, so a corrupt serialized context propagates the exception to user code.

## Offending code

`lib/src/api/otel_api.dart:608`

```dart
static TraceId traceIdFrom(String hexString) {
  _getAndCacheOtelFactory();
  try {
    final bytes = IdGenerator.hexToBytes(hexString);
    if (bytes == null || bytes.length != TraceId.traceIdLength) {
      throw const FormatException(
          'TraceId must be ${TraceId.traceIdLength} bytes');
    }
    return OTelFactory.otelFactory!.traceId(bytes);
  } catch (e) {
    throw FormatException('Invalid TraceId hex string: $hexString, $e');
  }
}
```

`spanIdFrom` at `lib/src/api/otel_api.dart:643` is the same shape.

## How to correct it

Return the safe default and report, instead of throwing:

```dart
static TraceId traceIdFrom(String hexString) {
  _getAndCacheOtelFactory();
  final bytes = IdGenerator.hexToBytes(hexString);
  if (bytes == null || bytes.length != TraceId.traceIdLength) {
    OTelErrorHandling.report(
        FormatException('Invalid TraceId hex string: $hexString'));
    return traceIdInvalid();
  }
  return OTelFactory.otelFactory!.traceId(bytes);
}
```

Same for `spanIdFrom` with `spanIdInvalid()`. This is the pattern `TraceState.put` already uses for a bad entry (`trace_state.dart:80`).

Tests must assert the report actually fires, not only that the return value is invalid. Install a handler and check the error arrives, as `otel_error_handler_test.dart` does:

```dart
final received = <Object>[];
OTelErrorHandling.handler = (error, _) => received.add(error);
expect(OTelAPI.traceIdFrom('4BF92F3577B34DA6A3CE929D0E0E4736').isValid, isFalse);
expect(received.single, isA<FormatException>());
```

Update the existing `throwsA(isA<FormatException>())` expectations in `trace_id_test.dart` and `span_id_test.dart` accordingly.

**BREAKING**: callers that catch `FormatException` from these methods must switch to checking `isValid`.

## Notes

- Spec version: **v1.60.0**.
- `SpanContext.fromJson` calls both methods. After the fix a corrupt id in the JSON yields an invalid `SpanContext` rather than an exception.
- The SDK propagator already checks `traceId.isValid` / `spanId.isValid` after the parse, so it is correct without changes once this lands; its `try/catch` becomes dead code for this path.
- Open question for the SDK half: a malformed inbound `traceparent` would now reach `OTelErrorHandling` on every request, where today it is a debug log. W3C says an invalid header is to be *ignored*, which is not an error condition. The SDK may want to validate the field grammar (`^[0-9a-f]{32}$`, `^[0-9a-f]{16}$`) before calling into the API so the error handler fires only on genuine misuse.
- `traceIdOf(Uint8List)` and `spanIdOf(Uint8List)` throw `ArgumentError` on a wrong length — the same violation. Out of scope here; either widen this issue or open a sibling.

i-am-paradox:
Fixes #132 using the approach you outlined in the issue: report through OTelErrorHandling and return the invalid id, same as TraceState.put does for a bad entry.

I updated the existing throws-expectations in trace_id_test and span_id_test, and added coverage that a handler-installed test actually receives the FormatException, including the wrong-length case. The full suite passes (1175 tests) and analyze is clean.

### DIFF
--- a/lib/src/api/otel_api.dart
+++ b/lib/src/api/otel_api.dart
@@ -604,18 +604,20 @@ class OTelAPI {
     return OTelFactory.otelFactory!.traceId(traceId);
   }
 
-  /// Creates a new [TraceId] from a hex string
+  /// Creates a new [TraceId] from a hex string.
+  ///
+  /// If the string is malformed or has the wrong length, the error is
+  /// reported to the error handler and an invalid [TraceId] is returned,
+  /// following the OpenTelemetry error-handling spec.
   static TraceId traceIdFrom(String hexString) {
     _getAndCacheOtelFactory();
-    try {
-      final bytes = IdGenerator.hexToBytes(hexString);
-      if (bytes == null || bytes.length != TraceId.traceIdLength) {
-        throw FormatException('TraceId must be {$TraceId.traceIdLength} bytes');
-      }
-      return OTelFactory.otelFactory!.traceId(bytes);
-    } catch (e) {
-      throw FormatException('Invalid TraceId hex string: $hexString, $e');
+    final bytes = IdGenerator.hexToBytes(hexString);
+    if (bytes == null || bytes.length != TraceId.traceIdLength) {
+      OTelErrorHandling.report(
+          FormatException('Invalid TraceId hex string: $hexString'));
+      return traceIdInvalid();
     }
+    return OTelFactory.otelFactory!.traceId(bytes);
   }
 
   /// Creates an invalid [Trace] (all zeros)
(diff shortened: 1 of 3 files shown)


michaelbushe:
Code changed underneath this.  I resolved one conflict (to your change).

michaelbushe (changes requested):
After merge resolution, six tests fail, all the ones Robert added in #112 that assert the throw you're removing:

trace_id_test.dart: rejects a correctly sized uppercase hex string
trace_id_test.dart: rejects a signed hex string instead of corrupting the id
trace_id_test.dart: rejects lowercase hex that is not 32 characters
span_id_test.dart:  rejects a correctly sized uppercase hex string
span_id_test.dart:  rejects a signed hex string instead of corrupting the id
span_id_test.dart:  rejects lowercase hex that is not 16 characters
  
It's all the same test pattern you already used: capture a handler, assert the returned id is invalid, assert one FormatException arrived. The inputs are worth keeping exactly as they are, since each one documents a specific way the old int.tryParse corrupted an id.

Leave traceIdOf and spanIdOf alone. They take raw bytes, still throw ArgumentError, and this PR doesn't touch them.

Two other things while you're in there. Six tests install an OTelErrorHandling handler and only two have addTearDown(OTelErrorHandling.resetToDefault). Without it a failing expectation leaks the capturing handler into every later test in the file, so one real failure becomes a cascade. 

Also, this needs a CHANGELOG entry under ### Fixed in 1.0.0-rc.4-wip; it changes behavior callers depend on, and the convention here is to link the PR rather than the issue.

There's also an analyze fix.  If you install the git-hooks (See CONTRIBUTING.md), it will be on pre-push.

Thank you.

PR #138, opened 2026-09-11
APISpan honors the isRecording creation parameter

### LINKED ISSUE #96: [BUG] Trace API: APISpan ignores the isRecording creation parameter

i-am-paradox:
Fixes #96. APISpan now stores the isRecording value given at creation and gates every mutating operation on it, so mutations on a non-recording span are no-ops.

One thing I noticed while wiring this: createSpan was ANDing the isRecording parameter with APITracer.isEnabled, which returns false on the base class, so every span created through an SDK-like factory was silently non-recording. I left isEnabled alone since the SDK overrides it as an optimization hint, and made the creation decision the single gate, which is what the existing tracer tests already expect.

Added coverage for both a false and a true creation decision in span_coverage_test. Full suite passes (1177 tests) and analyze is clean.

### DIFF
--- a/lib/src/api/otel_api.dart
+++ b/lib/src/api/otel_api.dart
@@ -604,18 +604,20 @@ class OTelAPI {
     return OTelFactory.otelFactory!.traceId(traceId);
   }
 
-  /// Creates a new [TraceId] from a hex string
+  /// Creates a new [TraceId] from a hex string.
+  ///
+  /// If the string is malformed or has the wrong length, the error is
+  /// reported to the error handler and an invalid [TraceId] is returned,
+  /// following the OpenTelemetry error-handling spec.
   static TraceId traceIdFrom(String hexString) {
     _getAndCacheOtelFactory();
-    try {
-      final bytes = IdGenerator.hexToBytes(hexString);
-      if (bytes == null || bytes.length != TraceId.traceIdLength) {
-        throw FormatException('TraceId must be {$TraceId.traceIdLength} bytes');
-      }
-      return OTelFactory.otelFactory!.traceId(bytes);
-    } catch (e) {
-      throw FormatException('Invalid TraceId hex string: $hexString, $e');
+    final bytes = IdGenerator.hexToBytes(hexString);
+    if (bytes == null || bytes.length != TraceId.traceIdLength) {
+      OTelErrorHandling.report(
+          FormatException('Invalid TraceId hex string: $hexString'));
+      return traceIdInvalid();
     }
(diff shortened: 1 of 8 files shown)


harshitt13:
Welcome, and thanks for the PR @i-am-paradox. Cross-linking #124 for visibility, which covers the same issue and has been open since Sep 2. So I will leave it to @michaelbushe and @robert-northmind to decide which one to take.

Worth noting separately: eb8cf24 on this branch fixes #132, not #96, and currently conflicts with main.

michaelbushe:
Thanks for this, and sorry you spent the time. I merged #124 instead, which covers the same issue and has been open since 2 September with some back and forth already.

It was slightly better in one line. Both found the same coupling: once APISpan honors the creation flag, createSpan's (isRecording ?? true) &&  isEnabled(...) starts mattering, and the base isEnabled hardcoded to false makes every span non-recording. Your diagnosis of that was exactly right, and you're the only one who wrote it down clearly.
 
The difference is the fix, let me know if you disagree. You removed the && isEnabled(...) conjunct and made the creation decision the only gate. #124 kept the conjunct and fixed the hardcoded false instead. trace/sdk.md says Enabled MUST return false when there are no registered span processors, so it's a correctness gate rather than an optimization hint, and a tracer that MUST report disabled shouldn't still be producing recording spans. That's why the conjunct stays.
 
Your _modifiable routing through the getter was tidier than #124's and the comments were better. Not enough to outweigh the spec point, but I appreciate the quality.

One process thing so this doesn't happen again: #96 was assigned to @harshitt13 and had a linked PR on it since 2 September. Both show on the issue page. Worth checking the assignee and the linked pull requests before starting.
 
However, most of the fault here is mine. #124 sat eleven days with no comment at all, which makes an assigned issue look abandoned. Your instinct to pick up something that looked dead was reasonable.  Thank you.

Issue #101, opened 2026-08-23
Tracking: OpenTelemetry spec-compliance audit, wave 2 — prioritized worklist


michaelbushe:
> Successor to #44, the July wave (closed — all its items landed). This wave holds the API-side findings from the August cross-model spec-compliance audit (see the SDK repo's shootout issue MindfulSoftwareLLC/dartastic_opentelemetry#95 and its worklist MindfulSoftwareLLC/dartastic_opentelemetry#256). Every entry below was verified against current `main` after the July fixes — this is what remains.

**Ordering rule:** the API is fixed first; SDK halves of split issues wait for the next API release. Do not work around a missing API fix in the SDK.

**How to claim:** check the box when you start (☑ + green chip = claimed/in progress, do not pick up; ☑ + purple = done; ☐ = up for grabs). If you can't edit this issue, comment on the individual issue to claim and a maintainer will tick it. Pick top-down within a tier.

## P0 — no-throw foundation & propagation primitives (12)

- [x] #94 — global error handler — **do this first**, later fixes route errors through it *(from SDK #232)*
- [ ] #80 — store empty string/array attribute values (they are meaningful) *(from SDK #233)*
- [ ] #81 — typed Attributes getters must not throw on type mismatch *(from SDK #234)*
- [ ] #82 — drop empty-key attributes, report, never throw *(from SDK #235)*
- [ ] #69 — addEvent must not throw on an empty event name *(from SDK #119)*
- [ ] #70 — API Meter must not validate instrument names or throw *(from SDK #141)*
- [x] #60 — Span.end() must not promote status Unset→Ok *(from SDK #107)*
- [x] #76 — CompositePropagator must extract in specified order *(from SDK #191)*
- [x] #77 — withSpanContext must derive a Context, never throw *(from SDK #196)*
- [ ] #91 — reject uppercase/signed hex in trace/span ID parsing (W3C) *(from SDK #193)*
- [ ] #92 — TraceFlags.fromString must not coerce invalid input to 0x00 *(from SDK #194)*
- [ ] #93 — TraceState parsing must enforce the W3C key/value grammar *(from SDK #195)*

## P1 — MUST completeness: scope fidelity, data model, API surface (21)

- [ ] #64 — getTracer/getLogger/getMeter must not invent scope version/schemaUrl *(from SDK #113)*
- [ ] #65 — span attributes must not become the Tracer scope attributes *(from SDK #114)*
[...]

PR #139, opened 2026-09-13
fix(api): stop throwing on empty instrument names and empty attribute keys

### LINKED ISSUE #70: [BUG] Metrics API: Meter validates instrument names and throws
> **Migrated from MindfulSoftwareLLC/dartastic_opentelemetry#141** — the fix belongs entirely in `dartastic_opentelemetry_api`; the SDK issue is closed as moved.

## Spec requirement

**Spec:** OpenTelemetry Specification v1.60.0 — Metrics API, Synchronous Instrument API
**Requirement level:** SHOULD NOT / MUST NOT
**URL:** https://opentelemetry.io/docs/specs/otel/metrics/api/#synchronous-instrument-api
**Pinned source:** https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/api.md#synchronous-instrument-api

> The API SHOULD be documented in a way to communicate to users that the `name` parameter needs to conform to the [instrument name syntax](https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-name-syntax). The API SHOULD NOT validate the `name`; that is left to implementations of the API, like the [SDK](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#instrument-name).

The No-Op spec raises this to MUST NOT for the API package, which runs as the no-op:

> the Meter MUST NOT validate any argument it receives.

> the MeterProvider MUST NOT return a non-empty error or log any message for any operations it performs.

## The problem

All seven instrument creation methods on the API meter check the name. Each one throws an `ArgumentError` when the name is empty.

The API package is the no-op implementation. Its own documentation says so. A no-op must accept every argument and must not validate anything.

An application that passes an empty name therefore crashes, and the crash comes from the layer that must stay silent.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/metrics/meter.dart:62`

```dart
if (name.isEmpty) {
  throw ArgumentError('Counter name must not be empty');
}
```

The same check exists in `createUpDownCounter` (`meter.dart:87`), `createHistogram` (`:115`), `createGauge` (`:142`), `createObservableCounter` (`:170`), `createObservableUpDownCounter` (`:199`), and `createObservableGauge` (`:228`).

## How to correct it

1. Remove the empty-name check and the throw from all seven `create*` methods.
2. Accept any name at the API level.
3. Document the name syntax in the doc comment. See #149.
4. Move validation to the SDK meter. There the SDK drops the instrument and writes a log. See #170.

## Notes

- Source finding files: `fable-ultra/metrics-api-01-name-validation-throws.md`, `fable-medium/metrics-api-instrument-name-validation-throws.md`, `grok/metrics/001-api-throws-empty-instrument-name.md`
- Parent issue: #95

### LINKED ISSUE #82: [BUG] Common: empty attribute keys are accepted without validation
> **Migrated from MindfulSoftwareLLC/dartastic_opentelemetry#235** — the fix belongs entirely in `dartastic_opentelemetry_api`; the SDK issue is closed as moved.

## Spec requirement

**Requirement level:** MUST

> An `Attribute` is a key-value pair, which MUST have the following properties:
>
> - The attribute key MUST be a non-`null` and non-empty string.

## The problem

No code path checks the attribute key. The constructor checks the value only.

The factory methods pass a key through unchecked as well.

`OTel.attributeString('', 'v')` therefore produces an attribute with an empty key. That attribute is stored and exported.

The non-null half of the rule holds, because Dart types the parameter as a non-nullable `String`. The non-empty half is unenforced.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/common/attribute.dart:23`

```dart
Attribute._(String key, T value)
    : _key = key,
      _value = value {
  if (_value is String && (_value as String).isEmpty) { ... }  // value checked
  // key is never checked
```

The unchecked paths are at `common/attribute_create.dart:16`, `factory/otel_api_factory.dart:170`, and `common/attributes.dart:33`.

## How to correct it

Drop an attribute whose key is empty. Write a warning with `OTelLog`. Do one check centrally, in the constructor or in the create method.

CAUTION: Do not throw. The error handling rules forbid it, and the empty-value issue in this same section removes the throws that already exist on this constructor.

## Notes

- Found by: Fable 5 (x-high run), Grok 4.5
- Source finding files: `fable-ultra/common-04-empty-attribute-key.md`, `grok/common/005-empty-attribute-keys.md`
- Parent issue: #95

michaelbushe:
Our last two P0 items from #101, both the "API MUST NOT throw" rule from [error-handling.md](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/error-handling.md#basic-error-handling-principles). One commit each so they can be reverted separately.

## Instrument names (#70)

All seven `APIMeter.create*` methods threw `ArgumentError` on an empty name. This is the no-op API; [metrics/noop.md](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/noop.md#meter) says the Meter MUST NOT return a non-empty error or log any message for any operation. The checks are removed and any name is accepted. No report either, for the same reason. Validation belongs to the SDK meter, which #170 there tracks.

The seven tests that asserted the throw now assert an instrument comes back and the error handler stays silent.

## Attribute keys (#82)

[common/README.md](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/common/README.md#attribute) says a key MUST be a non-empty string. Nothing checked it. Every `Attributes` is built through one private constructor, so the check lives there and covers `of`, `fromJson`, `attrsFromMap`, `attributesFromList` and every `copyWith` path. An empty-key attribute is dropped and reported through `OTelErrorHandling`. A bare `Attribute` with an empty key can still be constructed, but it cannot enter a collection, and the collection is the only place it was ever stored.

## Verification

- `dart analyze`: clean
- `dart test`: all pass
- Putting one Meter throw back fails the meter tests; removing the empty-key check fails the attributes tests, so both fixes are held by tests rather than accompanied by them.

Fixes #70
Fixes #82


robert-northmind (approved):
Looks good. nice work

PR #140, opened 2026-09-13
fix(api): APISpan no longer exposes what it recorded

### LINKED ISSUE #66: [BUG] Trace API: APISpan exposes attributes, events, links, and status

michaelbushe:
Fixes #66.

## What

`APISpan` exposed `attributes`, `spanEvents`, `spanLinks`, `status` and `statusDescription`. [trace/api.md, Span](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/trace/api.md#span):

> `Span`s are not meant to be used to propagate information within a process. To prevent misuse, implementations SHOULD NOT provide access to a `Span`'s attributes besides its `SpanContext`.

The `attributes` getter carried `@visibleForTesting`, which is an analyzer hint only. The getter was still public API, and a dynamic call reached it with no warning at all.

## How

The five accessors move to a `ReadableSpan`, reached through `getReadableSpan(span)`. Both are declared in a new part of `span.dart`, because the fields they read are library-private, and both are hidden from the package barrel:

```dart
export 'src/api/trace/span.dart'
    hide APISpanCreate, ReadableSpan, getReadableSpan;
```

Both carry `@internal`. An SDK reaches them by importing `src/api/trace/span.dart` directly and suppressing the lint at that one import, the same shape as the `// ignore: invalid_use_of_visible_for_testing_member` the SDK carries today, moved to one place.

The `attributes` **setter** is unchanged.

## Verification

- A barrel-only import can no longer reach them. I compiled a probe doing `span.attributes` against the public library and got `undefined_getter`.
- `dart analyze` clean, `dart test` 1231 passing.
- 127 call sites across 8 test files rewritten to `getReadableSpan(...)`.
- The web example no longer prints the span's status back out. Sample code should not teach the misuse the spec warns about.

## SDK impact

This breaks `dartastic_opentelemetry`'s `Span`, which re-exposes all five by delegation, as @harshitt13 found on #66. Tracked separately on the SDK side. The SDK is pinned below rc.4, so nothing breaks for users until that lands and the pin is widened.

## Severity

SDK users are unaffected in practice: the SDK's own `Span` gates every mutator on its independent recording state, so this reaches API-direct users and custom SDKs.


robert-northmind (approved):
Looks good. nice work!

PR #142, opened 2026-09-13
feat(api): add random-trace-id flag to TraceFlags

### LINKED ISSUE #104: [BUG] Trace API: TraceFlags has no random-trace-id flag (API half)
> **API half of MindfulSoftwareLLC/dartastic_opentelemetry#132** — fix the API first; the SDK issue stays open and tracks the SDK-side work (setting the flag on root span contexts), which depends on this landing in an API release.

## Spec requirement

**Spec:** W3C Trace Context Level 2 — Random Trace ID Flag
**Requirement level:** MUST (flag semantics), SHOULD (SDK sets it)
**URL:** https://www.w3.org/TR/trace-context-2/#random-trace-id-flag
**Pinned source:** https://www.w3.org/TR/trace-context-2/#trace-flags

> The second least significant bit of the trace-flags field denotes the random-trace-id flag.
>
> If that flag is set, at least the right-most 7 bytes of the trace-id MUST be selected randomly (or pseudo-randomly) with uniform distribution over the interval [0..2^56-1].
>
> If the flag is set in the incoming traceparent header, it MUST also be set in all outgoing traceparent headers which use the same trace-id.

The dependent SDK requirement is OpenTelemetry Specification v1.60.0 — Trace SDK, [Random trace flag](https://opentelemetry.io/docs/specs/otel/trace/sdk/#random-trace-flag):

> For root span contexts, the SDK SHOULD set the `Random` flag in the trace flags when it generates TraceIDs that meet the W3C Trace Context Level 2 randomness requirements.

## The problem

`TraceFlags` models one bit. It defines `NONE_FLAG = 0x0` and `SAMPLED_FLAG = 0x1`, and exposes `isSampled` and `withSampled`. There is no constant, accessor or setter for the random-trace-id bit (`0x02`).

The byte itself survives propagation — `asByte` carries all eight bits, so a `02` arriving on the wire is re-emitted as `02`. What is missing is the ability to *express* the bit:

- No `isRandom` getter, so nothing in the SDK, an instrumentation package or user code can ask whether an incoming trace ID is declared random. W3C §3.2.2.5.2 makes preserving the bit across a continued trace a MUST, and today that only works by accident of the whole byte being copied.
- No `withRandom`, so the SDK has no supported way to set it on a root span context. The `OTelFactory` pattern means `TraceFlagsCreate.create(0x02)` is not an option for callers outside the API package.

The randomness precondition is already satisfied. `IdGenerator` uses `Random.secure()` and fills all 16 bytes (`id_generator.dart:9,13-23`), so every trace ID from the default generator meets the "right-most 7 bytes uniformly random" rule.

## Offending code

`dartastic_opentelemetry_api/lib/src/api/trace/trace_flags.dart:16-21`

```dart
static const int NONE_FLAG = 0x0;
static const int SAMPLED_FLAG = 0x1;
```

`dartastic_opentelemetry_api/lib/src/api/trace/trace_flags.dart:57-66` — the single-bit accessor and setter, with no counterpart for the random bit:

```dart
bool get isSampled => (_flags & SAMPLED_FLAG) == SAMPLED_FLAG;

TraceFlags withSampled(bool isSampled) {
  if (isSampled) {
    return TraceFlagsCreate.create(_flags | SAMPLED_FLAG);
  } else {
    return TraceFlagsCreate.create(_flags & ~SAMPLED_FLAG);
  }
}
```

## How to correct it

1. Add `static const int RANDOM_FLAG = 0x02;` to `TraceFlags`.
2. Add `bool get isRandom => (_flags & RANDOM_FLAG) == RANDOM_FLAG;`.
3. Add `TraceFlags withRandom(bool isRandom)`, mirroring `withSampled`.
4. Document that the two bits are independent, and that the field is a bit set — a whole-value comparison such as `flags == sampled` is wrong once a second bit exists. W3C calls this out directly: `03` is both sampled and random.

The SDK half (MindfulSoftwareLLC/dartastic_opentelemetry#132) then sets the flag on root span contexts, and only for an ID generator that meets the randomness rule — see that issue's note and MindfulSoftwareLLC/dartastic_opentelemetry#131 for the generator hook.

## Notes

- Filed while sweeping the spec-compliance backlog: MindfulSoftwareLLC/dartastic_opentelemetry#132 names API-package files as its offending code but had no API-half issue, unlike every comparable dependency (api#83/#84/#85/#86/#87/#88/#89/#90/#91/#92/#93/#95/#99). This closes that gap.
- Related: `TraceFlags.fromString` leniency is tracked separately in #92.

harshitt13:
Refs #104

## What

Adds `RANDOM_FLAG` (`0x02`), `isRandom` and `withRandom` to `TraceFlags`, mirroring the existing sampled pair.

The flags byte already survived propagation via `asByte`, so an incoming `02` was re-emitted as `02` by accident of the whole byte being copied. What was missing was the ability to express the bit: nothing could ask whether an incoming trace ID is declared random, and the SDK had no supported way to set it on a root span context.

W3C Trace Context Level 2 makes preserving the bit across a continued trace a MUST. https://www.w3.org/TR/trace-context-2/#random-trace-id-flag

## Notes

- `withRandom` does not verify randomness and cannot. Setting the flag is an assertion by the caller; enforcing it belongs to the SDK, which only sets it for a conforming ID generator. Documented on the method.
- Class doc now states that `TraceFlags` is a bit set and that `flags == TraceFlags.sampled` is wrong once a second bit exists.
- Changelog entry under `Added`, not `Fixed`. Nothing changed for existing callers, so this is new public surface under Keep a Changelog. Per the conventions comment the spec citation carries the audit provenance, and #118 set the precedent of an audit item landing its additive surface under `Added`.
- No `TraceFlags.random` constant. The class doc argues against having named whole-value constants to compare against.
- Out of scope: `fromString` leniency (#92), and the mutable `none` and `sampled` statics, which I filed separately as #NNN.

## Follow-up: whole-value comparisons in tests

Grepping `== TraceFlags` and `equals(TraceFlags...)` turns up 9 sites. Only 4 should change.

**Wrong model**, assert "is sampled" via whole-value equality: `span_context_test.dart` lines 73, 119, 168, 169.

**Legitimate, leave alone**, round-trip assertions where the whole byte *is* the subject: `span_context_test.dart` lines 56, 254, 261, 281, and `span_context_create_test.dart:36`.

The criterion is what the assertion is about. A round trip on `withTraceFlags` asserts the setter stored the byte you handed it. A sampling check asserts a predicate, and equality only answers it while no other bit is set.

The clearest case is line 150, a test named `'trace flags indicates sampling state'` whose entire subject is the sampling predicate, tested by whole-value equality. Three of the four also carry a comment reading `// Use accessor method instead of isSampled property` directly above an `==` comparison, which has it backwards.

All 4 are inert today: they build their flags explicitly, so no random bit reaches them. They fail once root spans start emitting `03`. `git log -L` puts them in the initial revision, so it is original habit rather than a later sweep, which means the SDK repo likely carries the same idiom from the same authorship. That is the urgent half, since that is where #132 makes root spans `03` for real.

## Testing

Covers all four combinations of the two bits, both orders of application, that each `with*` preserves the other bit, and round trips through `fromString`/`toString`.


robert-northmind (approved):
I think it looks good. I am just thinking if we should shorten the docs a bit? 
They all seem a bit verbose. What do you think?

harshitt13:
> I think it looks good. I am just thinking if we should shorten the docs a bit? They all seem a bit verbose. What do you think?

Yeah, I will update it now.

robert-northmind (approved):
Looks good

